PR #2, opened 2021-01-11
Update readme.md with a snippet template

limonspb:


README.md
```diff
@@ -1,3 +1,24 @@
 # Firmware Presets
 
 Configuration Snippets for the Betaflight Flight Controller Firmware
+
+
+## Snippet template
+
+```
+# preset title
```
mikeller, line 9: How are these meant to be parsed? Based on line number? If so, would it be better to use a keyword based approach like `# Title: foo`? Or are they not meant to be parsed?

limonspb: @mikeller  yeah just line index I think is easier for everyone, hard to mess it up for those who will make presets, no special format.
line 1 - title
line 2 - BF version
all next lines - description until the empty line or line without # at the beginning is met.

mikeller: Not sure I'd agree to the 'hard to mess up' part. Especially not if more metadata is added at a later stage (think version _range_, MCU type, author, ...).
Is your parsing resistant to empty lines / empty comment lines sprinkled in between the first few lines?

limonspb: @mikeller ok it makes sense. What are your suggestions?
# Title: some title
# BetaflightVersion: 4.2
# Description: some text line 1
# Description: some text line 2
# Description: some text line 3
.....

Then should i check the entire file for these?

For now it just ALWAYS reads first line as a title. Second as a BF version. And all the next lines as a description untill there is a line without #

mikeller: The above looks good to me. I'd keep it as simple as possible while still giving you the features you need, so parse all comments in the file, and for single values use the first occurrence, for multilines (e.g. Description) aggregate all of them in order.

limonspb: @mikeller  what is the proper way for a script to report errors? For example, if there is no `# Title:` or multiple lines with `# Version`?

mikeller (approved):

PR #3, opened 2021-01-15
indexer script with flexible preset metadata

limonspb:
Format for presets metadata:

```
# Title: some title sample 1
# FirmwareVersion: 4.3 (optional)
# Author: Name (optional)
# Description: Description line1 (optional)
# Description: Description line2 (optional)
# Description: Description line3 (optional)
commands
```

Easy to add new metadata property to presets, just add a new item to this array in `indexer.js`:

```
const presetsFileMetadata = {
    title:             {value: "", optional: false  },
    firmwareVersion:   {value: "", optional: true },
    description:       {value: [], optional: true },
    ..........
}
```

Parser detects possible errors and the script returns "100" if there are any. 0 otherwise.
Example:

```
> node indexer/indexer.js
presets/TestSamples/4.3/sample1.txt: line 2, duplicated property title
presets/TestSamples/4.3/sample1.txt: line 17, duplicated property firmwareVersion
presets/TestSamples/4.3/sample2.txt: missing or empty title
```


indexer/PresetsFile.js
```diff
@@ -4,47 +4,109 @@ const crypto = require('crypto');
...
+                line = line.slice(lineBeginning.length).trim(); // (Title: foo) -> (foo)
+                this._processProperty(property, line);
             }
         }
+    }
 
-        delete this._remainingHeaders;
+    _processProperty(property, line)
+    {
+        if (Array.isArray(this._presetsFileMetadata[property].value)) {
```
mikeller, line 79: Interesting way of doing things. :-)

mikeller (approved):

indexer/PresetsFile.js
limonspb: ha! I just wanted so that it was easy and transparent how to add/remove presets metadata properties and configure them if a property is just a string or array :)

mikeller: We'll see how well this works once we want to add support for other data types. :wink:

PR #8, opened 2021-05-04
5in 4s race by ctzsznooze

ctzsnooze:
Suggestion for a 4S racer preset.
Assumes a reasonable radio link; not a lot of feedforward smoothing.
Launch control and radio settings aren't configured.

### DIFF
--- a/presets/TestSamples/4.3/ctzsnzooze_5in_4S_Race.txt
+++ b/presets/TestSamples/4.3/ctzsnzooze_5in_4S_Race.txt
@@ -5,7 +5,6 @@
 # Keywords: 5in, 4S, race
 # Author: Chris Thompson ctzsnooze
 # Description: Basic 5" 4S race setup for 4.3 by ctzsnooze
-# Description: WARNING: Disables Runaway Takeoff Prevention!
 # Description: Assumes a decent radio link.
 
 # General settings
@@ -15,7 +14,7 @@ set mixer_type = LEGACY
 set gyro_cal_on_first_arm = OFF
 set deadband = 0
 set yaw_deadband = 0
-set runaway_takeoff_prevention = OFF
+set runaway_takeoff_prevention = ON
 set thrust_linear = 0
 set transient_throttle_limit = 0
 set pwr_on_arm_grace = 1
--- a/presets/TestSamples/4.3/ctzsnzooze_5in_4S_Race.txt
+++ b/presets/TestSamples/4.3/ctzsnzooze_5in_4S_Race.txt
@@ -1,0 +1,140 @@
+# Title: 5" 4S racer ctzsznooze
+# FirmwareVersion: 4.3
+# Category: MIXED
+# Official: true
+# Keywords: 5in, 4S, race
+# Author: Chris Thompson ctzsnooze
+# Description: Basic 5" 4S race setup for 4.3 by ctzsnooze
+# Description: assumes a decent radio link
+
+# General settings
+
+set level_race_mode = OFF
+set mixer_type = LEGACY
+set gyro_cal_on_first_arm = OFF
+set deadband = 0
+set yaw_deadband = 0
+set runaway_takeoff_prevention = OFF
+set thrust_linear = 0
+set transient_throttle_limit = 0
+set pwr_on_arm_grace = 1
+set yaw_spin_recovery = AUTO
+set feedforward_transition = 0
+set frsky_vfas_precision = 1
+
+
+# Feedforward smoothing, assumes decent link
+
+set feedforward_averaging = NONE
+set feedforward_smooth_factor = 20
+set feedforward_jitter_reduction = 5
+set feedforward_max_rate_limit = 100
+set feedforward_boost = 13
+
+
+# Dynamic idle
+
+set dyn_idle_min_rpm = 25
+set dyn_idle_p_gain = 50
+set dyn_idle_i_gain = 50
+set dyn_idle_d_gain = 50
+set dyn_idle_max_increase = 150
+
+
+# Rates
+
+set rates_type = ACTUAL
+set roll_rc_rate = 10
+set pitch_rc_rate = 10
+set yaw_rc_rate = 10
+set roll_srate = 63
+set pitch_srate = 63
+set yaw_srate = 63
+set throttle_limit_type = SCALE
+set throttle_limit_percent = 95
+set tpa_rate = 65
+set tpa_breakpoint = 1350
+set tpa_mode = D
+
+
+# PIDs
+
+set p_pitch = 38
+set i_pitch = 81
+set d_pitch = 32
+set f_pitch = 105
+set p_roll = 35
+set i_roll = 77
+set d_roll = 30
+set f_roll = 100
+set p_yaw = 35
+set i_yaw = 80
+set d_yaw = 0
+set f_yaw = 80
+
+set d_min_roll = 22
+set d_min_pitch = 24
+set d_min_yaw = 0
+
+set d_min_boost_gain = 35
+set d_min_advance = 0
+
+set anti_gravity_mode = SMOOTH
+set anti_gravity_threshold = 250
+set anti_gravity_gain = 3000
+
+set iterm_relax_cutoff = 25
+
+
+# Battery and throttle
+
+set report_cell_voltage = ON
+
+set vbat_sag_compensation = 100
+set vbat_display_lpf_period = 10
+set vbat_sag_lpf_period = 2
+set vbat_sag_compensation = 100
+
+set throttle_boost = 5
+set throttle_boost_cutoff = 20
+
+
+# Filters
+
+set gyro_lowpass_type = PT1
+set gyro_lowpass_hz = 200
+set gyro_lowpass2_type = PT1
+set gyro_lowpass2_hz = 250
+set dyn_lpf_gyro_min_hz = 200
+set dyn_lpf_gyro_max_hz = 500
+set dyn_lpf_gyro_curve_expo = 5
+
+set gyro_notch1_hz = 0
+set gyro_notch1_cutoff = 0
+set gyro_notch2_hz = 0
+set gyro_notch2_cutoff = 0
+
+set dyn_lpf_dterm_min_hz = 70
+set dyn_lpf_dterm_max_hz = 170
+set dyn_lpf_dterm_curve_expo = 6
+set dterm_lowpass_type = PT1
+set dterm_lowpass_hz = 200
+set dterm_lowpass2_type = PT1
+set dyn_lpf_dterm_curve_expo = 6
+
+set dyn_notch_bandwidth_hz = 30
+set dyn_notch_count = 2
+set dyn_notch_min_hz = 160
+set dyn_notch_max_hz = 600
+
+set rc_smoothing = ON
+set rc_smoothing_auto_factor = 25
+set rc_smoothing_auto_factor_throttle = 20
+set rc_smoothing_setpoint_cutoff = 0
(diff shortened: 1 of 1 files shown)


presets/TestSamples/4.3/ctzsnzooze_5in_4S_Race.txt
```diff
@@ -0,0 +1,140 @@
...
+# Description: assumes a decent radio link
+
+# General settings
+
+set level_race_mode = OFF
+set mixer_type = LEGACY
+set gyro_cal_on_first_arm = OFF
+set deadband = 0
+set yaw_deadband = 0
+set runaway_takeoff_prevention = OFF
```
etracer65, line 17: Silently removing a critical safety feature is not acceptable.

etracer65 (changes requested):

presets/TestSamples/4.3/ctzsnzooze_5in_4S_Race.txt
ctzsnooze: Good point.  Some race tracks have very sudden turns after takeoff and it can be difficult to ensure that runaway prevention doesn't trip when it shouldn't, so i have it off on all my race setups.
I'll include a comment.

etracer65: I didn't mean add a comment and leave it in. I meant to not disable a safety feature. A comment about how it can be disabled in the unlikely event of a problem is ok. But that comment should not encourage disabling or exaggerate the risk. The feature has been enabled for millions of flights and pretty much for every race lap flown without problem.

ctzsnooze: OK... in that case, what values would you recommend so for a racer that turns hard nearly immediately after takeoff, so that the runaway protection disables it self really quickly and reliably?

McGiverGim: I'm with @etracer65 on this. We don't must remove any safety feature with presets. The user must disable it conscious and knowing what he is doing.

McGiverGim:
Another thing that is strange to me, is that this "presets" include a lot of different things, from PID/filter adjustment to rates, disable gyro calibration, some frsky adjustment... I don't know if the presets are accumulative, if that is true maybe they need to be more "basic". One preset for adjusting PIDs and filters, another one for different rates, etc. and the final user can decide how many of them apply.

I can't see a real utility to all this presets if they have too much changes at once, and the major part of users will need to change/customize them.

But this is more a general question about where want we go with this presets.

presets/TestSamples/4.3/ctzsnzooze_5in_4S_Race.txt
etracer65: > OK... in that case, what values would you recommend so for a racer that turns hard nearly immediately after takeoff, so that the runaway protection disables it self really quickly and reliably?

I don't think any settings changes are needed. It's actually a fallacy that a hard move immediately after takeoff will trigger runaway takeoff prevention. The feature works based on the pidsum growing very large for a longish period of time. The pidsum must exceed 600 (60.0 as displayed in blackbox) and stay above that limit for 75ms. A quad must be poorly tuned for this to be inadvertently triggered. I would assume that the entire point of a "racer" preset is that it contains settings to make the pid controller track setpoint well and thus not have a concern with large and persistent error. Further the code contains logic to further accelerate deactivation under high throttle conditions as you would expect early after a race launch.

presets/TestSamples/4.3/ctzsnzooze_5in_4S_Race.txt
```diff
@@ -0,0 +1,140 @@
...
+
+set level_race_mode = OFF
+set mixer_type = LEGACY
+set gyro_cal_on_first_arm = OFF
+set deadband = 0
+set yaw_deadband = 0
+set runaway_takeoff_prevention = ON
+set thrust_linear = 0
+set transient_throttle_limit = 0
+set pwr_on_arm_grace = 1
```
etracer65, line 20: This is another safety feature that shouldn't be changed.

etracer65 (changes requested):

etracer65:
In general I think the presets shouldn't contain any "preference" items. They should only include flight performance types settings. So even settings that affect rates should not be here.

Also the preset should not contain lines that simply set defaults. Like:
```
 set level_race_mode = OFF
 set mixer_type = LEGACY
 set gyro_cal_on_first_arm = OFF
 set deadband = 0
 set yaw_deadband = 0
 set runaway_takeoff_prevention = ON
 set thrust_linear = 0
 set transient_throttle_limit = 0
```
and others. It's the default so it doesn't need to be here. And it's dangerous that it is because later if a default changes in the firmware then the preset will override this. The same rules apply to unified targets - no default settings included.

limonspb:
> In general I think the presets shouldn't contain any "preference" items. They should only include flight performance types settings. So even settings that affect rates should not be here.
> 
> Also the preset should not contain lines that simply set defaults. Like:
> 
> ```
>  set level_race_mode = OFF
>  set mixer_type = LEGACY
>  set gyro_cal_on_first_arm = OFF
>  set deadband = 0
>  set yaw_deadband = 0
>  set runaway_takeoff_prevention = ON
>  set thrust_linear = 0
>  set transient_throttle_limit = 0
> ```
> 
> and others. It's the default so it doesn't need to be here. And it's dangerous that it is because later if a default changes in the firmware then the preset will override this. The same rules apply to unified targets - no default settings included.

limonspb:
> In general I think the presets shouldn't contain any "preference" items. They should only include flight performance types settings. So even settings that affect rates should not be here.
> 
> Also the preset should not contain lines that simply set defaults. Like:
> 
> ```
>  set level_race_mode = OFF
>  set mixer_type = LEGACY
>  set gyro_cal_on_first_arm = OFF
>  set deadband = 0
>  set yaw_deadband = 0
>  set runaway_takeoff_prevention = ON
>  set thrust_linear = 0
>  set transient_throttle_limit = 0
> ```
> 
> and others. It's the default so it doesn't need to be here. And it's dangerous that it is because later if a default changes in the firmware then the preset will override this. The same rules apply to unified targets - no default settings included.

Agree. For defaults, we need to use `# defaults` meta property where you can specify the snippet file for defaults. (which we haven't prepared yet for tuning)
This preset is more like from "mixed" category if it includes rates and other personal stuff.

ctzsnooze:
There are a number of issues here that go to the heart of a 'preset'.
To me what matters is that a user can achieve a particular goal.
Let's say the user wants the quad to be 'smoother on the sticks', for a more cinematic feel.  We can do this with a small preset that modifies feedforward, rc_smoothing, and not much else.  They don't want a re-tune, or to change anything other than stick feel.  We could therefore provide several 'stick feel' variants, from 'super-sharp' through to 'cinematic'.
In this kind of example, we aren't wanting to change their tune, or alter their anti-gravity settings, or OSD display layout.
A simple, very limited preset is all that's needed to do that.
And there are lots of similar 'partial' presets, for example something to configure an OSD a certain way, or set up a certain type of VTx,, or, perhaps, configure an Rx link, or an LED layout, etc.
But let's say the user has made all kinds of changes, perhaps un-intentionally.  And they want to put a preset in place that works exactly like some other user's race setup, for example.  In this case, if they had accidentally changed thrust linear to be on, but the racer in question always had it off, then the preset *should* contain thrust linear = off.  Even though that is the default.
Hence it makes sense in a 'complete' preset to confirm that defaults that should be off, are off.
And of course the dilemma is what to include and what to exclude.
In the example above, I deliberately force some values 'back to' defaults, because in my experience, they needed to be set to the default value for the preset to work as intended.

ctzsnooze:
eTracer: "It's the default so it doesn't need to be here. And it's dangerous that it is because later if a default changes in the firmware then the preset will override this."
I don't agree with the idea that its dangerous for a preset to enforce a default.  
Its certainly no dangerous than enforcing a non-default value.  Both are equally hazardous.
The preset model here is that each preset is defined for the firmware level that it is specified for, using the appropriate tags and with information in the headers.  If within one firmware release version, a parameter gets changed so much that the preset needed editing for safety reasons, it will be our task to annotate t the original version of the preset should to indicate that it should not be used for versions higher than x.x, and create a new version for x.x and above that has values appropriate for the newer version, if that can be done.
I think we need to realise that maintaining presets will be difficult and time consuming if there are significant changes within a version.
When a new version of betaflight comes out, there will be a need to check out if each preset requires some adaptation to suit the new firmware, or if it can be retained 'as is' with only the tag requiring updating.

ctzsnooze:
Reasons:
set level_race_mode = OFF ## some people turn it on, for real racing it should be off
 set mixer_type = LEGACY ## linear interpolation used to cause spikes
 set gyro_cal_on_first_arm = OFF ## to avoid delay if someone is late arming at the start
 set deadband = 0 ## in case the user turned it on and forgot
 set yaw_deadband = 0 ## in case the user turned it on and forgot
 set runaway_takeoff_prevention = ON ## because racers often put props on wrong
 set thrust_linear = 0 ## because for race motor setups I think its better off
 set transient_throttle_limit = 0 ## because it should be off

etracer65:
Doing this will come back and bite us in the ass later. I will not support presets containing settings which are default values.

In general I think the entire "presets" concept is doomed to failure just like having the "community presets" on the wiki. The only way it's viable is to have very generalized and simplified safe settings that help push the user into the correct direction for a particular flight style and quad type. As soon as they get specific with settings they are doomed to fail.

Also consider that the only people that will really use this "race" preset are people who really aren't racers. A real racer with experience tuning his quad isn't going to use a preset. Beginners will try this thinking it will make them "fast". And since they're least likely to be able to understand how to tune or even how the quad should feel being extreme on the setup is likely to result in a bad experience.

limonspb:
I guess for each preset we need to figure out the purpose.
I see there can be two different ones: one that @etracer65  talks about and the other @ctzsnooze  talks about.

1. The first ones are mainly for beginners in tuning. They just build "8 inch long range" or "3 inch 4S" and they just want to make sure it won't freak out and will have a decent tune, like 80-90% tune and that will be more than enough for them. They click apply and that's it. Sometimes these people tend to mess up the tune and make things worse following stupid videos on youtube. So for this presets category it is important to use # defaults and apply the actual preset on top of that. These presets should **not** have any default values as all the default values will be in defaults_tune_4.3.txt
2. The Second category is the one @ctzsnooze is talking about: a bit more complicated, specific and precise, maybe for a little more experienced users. They might apply RC filtering or some tuning on top of what the user already have.
3. The third additional category is probably some presets that guys like @ctzsnooze just want to share. They will be mixed category and can include any personal preferences. I would not mark them as official, but that's debatable.

(not talking about presets for BNF/RTF - they are pretty straightforward)

ctzsnooze:
I'm happy with this preset now and am removing the draft flag.
It configures the things that I think are important for a tune of this kind, both actively by setting a value, and by forcing some settings back to defaults.
Very keen to get some user feedback.

limonspb:
nice! Why do you have both pid values and sliders?

ctzsnooze:
Yes, this is ready I think

ctzsnooze:
Indeed, and precisely why this TUNE has all it needs.  
It will be fine when flashed over anything.  
There is no 'dirty tune' possibility here because I have controlled the values that matter.

limonspb:
@ctzsnooze think about presets that people will apply after your tune. There should be no leftovers after your presets. No throttle curve, no throttle limit and no other weird settings.
The only way to do that is to allow only setrain CLI in TUNE category. Specify this list, add it to defaults.txt and strictly follow it.

PR #9, opened 2021-05-11
Skip .* files/folders and non txt files for presets indexing

limonspb:
After discussion with @ctzsnooze figured we need to skip .* files and folders while indexing presets.
Also decided to skip all non-txt files to be safe.


ctzsnooze (approved):
Good

ctzsnooze (commented):
great, important for mac in particular

PR #10, opened 2021-05-19
Default 4.3 rates and BMSThomas rates

limonspb:
Thomas "BMSThomas" Bitmatta - MultiGP IO 2019 Champion.
image


ctzsnooze:
I think that 'rates' should mean only rates (the deg/s stuff), not all things that happen to be stored in the rates profile.
So if a preset is titled "Fred's rates" it should not include throttle curve.
If it is titled "Fred's rates and throttle curve" that's OK.

ctzsnooze (dismissed):

limonspb:
removed throttle curve and throttle limits from rates

ctzsnooze (approved):

PR #11, opened 2021-05-19
Allow multiple default snippets for a preset

limonspb:
This will allow using of multiple defaults like that:

```
# Title: Preset title
# FirmwareVersion: 4.2
# FirmwareVersion: 4.3 (as many versions as needed)
# Category: TUNE
# Official: true
# Keywords: word1, word2, word3
# Author: Name Lastname Pilotname
# Description: Description line1
# Description: Description line2  (as many description lines as needed)
# Defaults: path/to/defaults1.txt
# Defaults: path/to/defaults2.txt (as many default snippets as needed)

<cli command 1>
<cli command 2>
....
<cli command n>
```


etracer65:
The defaults are already contained in the firmware. There is no reason for them to be in the preset or applied in any alternative manner. Just a simple `defaults nosave` is all that's needed. Trying to maintain sets of defaults outside the firmware just adds extra work and increases the chances of mistakes or discrepancies. Presets should only contain changes to the default settings.

limonspb:
@etracer65 can we reset everything for a person that applies OSD, TUNE or VTX presets? What if he already has setup ports, modes, failsafe and something else. When he uses "**TUNE 5 inch racer**" preset, he expects only the tune to be affected.
And we still have to reset all tune parameters in case he messed up his tune already.

I agree it is extra work of creating `defaults_tune_4.3.txt`, `defualts_rates_4.3.txt` etc.
But it's only 9 files for each release (9 categories). So basically 9 simple files once per year:
`TUNE, RATES, OSD, VTX, LEDS, MODES, BNF, RC_SMOOTHING, MIXED`
and its quite easy to create them - copy paster parts of the default dump.

etracer65:
If you're concerned about the user not having some settings to defaults and thus affecting the "tune", then the instructions should be that they reset their settings to defaults before applying the preset. There are many settings that might affect the "tune", how could the snippet possibly have all of them included? I'm sure I could change many settings that will break the tune that you wouldn't have included in your "defaults". Take an extreme example, the user has disabled the dynamic filter feature. Are you going to make sure every tuning related preset has `feature DYNAMIC_FILTER`. What if they have the PID frequency less than 2K and this automatically disables the dynamic filter? What then? What about if the user changed some setting that ends up disabling rpm filtering? Of maybe even as simple as enabled D on yaw? Are you going to put `set d_yaw = 0` in every preset? If you're preset is relying on the user having defaults, then make that a requirement and use the mechanism that already exists for defaults. The point is that you can't account for everything with your "shadow" defaults so it's a losing proposition to duplicate them in yet another non-maintainable way.

I would suggest that the preset have a flag that determines when a "reset to defaults" is required. Presets that are vulnerable to the user's possibly incompatible settings like extensive tuning should have this set to "yes". Presets that maybe only adjust rates or are less inclusive might not need to make this a requirement. When a user tries to apply a preset that requires it then they can be warned and asked if they want to continue.

limonspb:
> If you're concerned about the user not having some settings to defaults and thus affecting the "tune", then the instructions should be that they reset their settings to defaults before applying the preset. There are many settings that might affect the "tune", how could the snippet possibly have all of them included? I'm sure I could change many settings that will break the tune that you wouldn't have included in your "defaults". Take an extreme example, the user has disabled the dynamic filter feature. Are you going to make sure every tuning related preset has `feature DYNAMIC_FILTER`. What if they have the PID frequency less than 2K and this automatically disables the dynamic filter? What then? What about if the user changed some setting that ends up disabling rpm filtering? Of maybe even as simple as enabled D on yaw? Are you going to put `set d_yaw = 0` in every preset? If you're preset is relying on the user having defaults, then make that a requirement and use the mechanism that already exists for defaults. The point is that you can't account for everything with your "shadow" defaults so it's a losing proposition to duplicate them in yet another non-maintainable way.
> 
> I would suggest that the preset have a flag that determines when a "reset to defaults" is required. Presets that are vulnerable to the user's possibly incompatible settings like extensive tuning should have this set to "yes". Presets that maybe only adjust rates or are less inclusive might not need to make this a requirement. When a user tries to apply a preset that requires it then they can be warned and asked if they want to continue.

@etracer65  the problem is that there is no way for a user to reset TUNE without affecting other settings (modes, ports, OSD etc). Weird settings like 2k update rate or enabled d_yaw is exactly a reason why we need **defaults_4.3_tune.txt**. It will reset all tune parameters to defaults and by all i mean ALL tune-related parameters. It is not hard to create this file once we finalize 4.3, just copy-paste parts of the dump. It is an extra hour of work for us per year only. But simplifies the life of the user by a lot. I think for presets it is unacceptable to even ask the user if he wants to reset **everything**. It will just make presets much less attractive.

Typical use case: someone built a 8 inch drone. He can arm it and it flies but flies like crap. He asked his friend to tune and his friend failed :) followed youtube - failed as well. Now he wants to try presets. We can't reset everything but we have to reset his **tune** before applying "8 inch 6S" preset.

etracer65:
You're fooling yourself if you think you can cherry-pick only the defaults you need to reset and no other changes set by the user will affect the end result. Further you're creating a maintenance headache for the project by duplicating the defaults. Take the BF4.0 situation. Because of overly aggressive settings in the initial release 4.0.1 needed new tuning defaults to avoid flyaways. So are you going to maintain defaults all the way to the maintenance release granularity?

As I said before - I will not support default settings being made parts of presets. It's a bad idea, and bad for the future of the project. Either the preset needs to have default settings to work properly or it doesn't. Trying to arbitrarily decide which settings are "okay" to retain and which aren't is subjective and will lead to constant confusion.

limonspb:
> You're fooling yourself if you think you can cherry-pick only the defaults you need to reset and no other changes set by the user will affect the end result. Further you're creating a maintenance headache for the project by duplicating the defaults. Take the BF4.0 situation. Because of overly aggressive settings in the initial release 4.0.1 needed new tuning defaults to avoid flyaways. So are you going to maintain defaults all the way to the maintenance release granularity?
> 
> As I said before - I will not support default settings being made parts of presets. It's a bad idea, and bad for the future of the project. Either the preset needs to have default settings to work properly or it doesn't. Trying to arbitrarily decide which settings are "okay" to retain and which aren't is subjective and will lead to constant confusion.

Seems like unfortunately there is no easy way here. Either user struggles or we maintain presets defaults every time default changes. The last one isn't happening too often, once per year, isn't it?

It's pretty simple to cherry-pick defaults for rates, for vtx tables, for OSD, right? What about TUNE... Not so easy, i get it, its' just much more parameters. 4.3 not ready, so lets check 4.2 quickly. Took me 10 minutes to get that. lets discuss what you don't agree with in these TUNE defaults:

```
# feature
feature AIRMODE
feature ANTI_GRAVITY
feature DYNAMIC_FILTER

# mixer
mixer QUADX
mmix reset

# master
set gyro_hardware_lpf = NORMAL
set gyro_lowpass_type = PT1
set gyro_lowpass_hz = 200
set gyro_lowpass2_type = PT1
set gyro_lowpass2_hz = 250
set gyro_notch1_hz = 0
set gyro_notch1_cutoff = 0
set gyro_notch2_hz = 0
set gyro_notch2_cutoff = 0
set gyro_calib_duration = 125
set gyro_calib_noise_limit = 48
set dyn_notch_width_percent = 8
set dyn_notch_q = 120
set dyn_notch_min_hz = 150
set dyn_notch_max_hz = 600
set dyn_lpf_gyro_min_hz = 200
set dyn_lpf_gyro_max_hz = 500
set dyn_lpf_gyro_curve_expo = 5
set acc_lpf_hz = 10
set airmode_start_throttle_percent = 25
set min_throttle = 1070
set max_throttle = 2000
set min_command = 1000
set dshot_idle_value = 550
set dshot_burst = ON
set dshot_bidir = OFF
set motor_pwm_rate = 480
set motor_poles = 14
set mixer_type = LEGACY
set thrust_linear = 0

# profiles
set dyn_lpf_dterm_min_hz = 70
set dyn_lpf_dterm_max_hz = 170
set dyn_lpf_dterm_curve_expo = 5
set dterm_lowpass_type = PT1
set dterm_lowpass_hz = 150
set dterm_lowpass2_type = PT1
set dterm_lowpass2_hz = 150
set dterm_notch_hz = 0
set dterm_notch_cutoff = 0
set vbat_sag_compensation = 0
set pid_at_min_throttle = ON
set anti_gravity_mode = SMOOTH
set anti_gravity_threshold = 250
set anti_gravity_gain = 3500
set feedforward_transition = 0
set iterm_rotation = OFF
set iterm_relax = RP
set iterm_relax_type = SETPOINT
set iterm_relax_cutoff = 15
set iterm_windup = 100
set iterm_limit = 400
set pidsum_limit = 500
set pidsum_limit_yaw = 400
set yaw_lowpass_hz = 0
set throttle_boost = 5
set throttle_boost_cutoff = 15
set p_pitch = 46
set i_pitch = 90
set d_pitch = 38
set f_pitch = 95
set p_roll = 42
set i_roll = 85
set d_roll = 35
set f_roll = 90
set p_yaw = 45
set i_yaw = 90
set d_yaw = 0
set f_yaw = 90
set angle_level_strength = 50
set horizon_level_strength = 50
set horizon_transition = 75
set level_limit = 55
set horizon_tilt_effect = 75
set horizon_tilt_expert_mode = OFF
set abs_control_gain = 0
set abs_control_limit = 90
set abs_control_error_limit = 20
set abs_control_cutoff = 11
set use_integrated_yaw = OFF
set integrated_yaw_relax = 200
set d_min_roll = 23
set d_min_pitch = 25
set d_min_yaw = 0
set d_min_boost_gain = 37
set d_min_advance = 20
set ff_interpolate_sp = ON
set ff_max_rate_limit = 100
set ff_smooth_factor = 37
set ff_boost = 15
set dyn_idle_min_rpm = 0
set dyn_idle_p_gain = 50
set dyn_idle_i_gain = 50
set dyn_idle_d_gain = 50
set dyn_idle_max_increase = 150



set tpa_rate = 65
set tpa_breakpoint = 1350
set tpa_mode = D
 
```

etracer65:
I'm not going to spend much time on this since as I said I don't support defaults in presets. But even your example demonstrates the pitfalls.
```
set gyro_calib_duration = 125
set gyro_calib_noise_limit = 48
```
Why would these be part of a "tune". Why reset the settings required to get my noisy gyro to calibrate?
```
set dshot_idle_value = 550
```
Why change the idle of my motors? This is based on the ESCs and motors I use - not the tune.
```
set dshot_burst = ON
```
This will make DSHOT not work on half the flight controllers
```
set dshot_bidir = OFF
```
So now we're discouraging people from using rpm filtering?
```
set motor_poles = 14
```
I guess it doesn't matter since rpm filtering is off, but if I turn it on my motor poles may be wrong now.

So that's enough after spending just 30 seconds looking at the list. So as soon as you start playing games cherry-picking which "defaults" you think need to be reset then the user is left wondering which of the settings the need (or want) to use are now no longer valid. So either tell the user up front that all settings will be reset as a first step in using this preset, or don't reset any settings.

ctzsnooze (approved):
Looks great!

etracer65:
@ctzsnooze This is unacceptable! Merging a pull request that was opened today and is obviously under contention is an abuse of your privileges on the project! This should be grounds for removal from the project.

limonspb:
> I'm not going to spend much time on this since as I said I don't support defaults in presets. But even your example demonstrates the pitfalls.
> 
> ```
> set gyro_calib_duration = 125
> set gyro_calib_noise_limit = 48
> ```
> 
> Why would these be part of a "tune". Why reset the settings required to get my noisy gyro to calibrate?
> 
> ```
> set dshot_idle_value = 550
> ```
> 
> Why change the idle of my motors? This is based on the ESCs and motors I use - not the tune.
> 
> ```
> set dshot_burst = ON
> ```
> 
> This will make DSHOT not work on half the flight controllers
> 
> ```
> set dshot_bidir = OFF
> ```
> 
> So now we're discouraging people from using rpm filtering?
> 
> ```
> set motor_poles = 14
> ```
> 
> I guess it doesn't matter since rpm filtering is off, but if I turn it on my motor poles may be wrong now.
> 
> So that's enough after spending just 30 seconds looking at the list. So as soon as you start playing games cherry-picking which "defaults" you think need to be reset then the user is left wondering which of the settings the need (or want) to use are now no longer valid. So either tell the user up front that all settings will be reset as a first step in using this preset, or don't reset any settings.

On a side note - `# defaults` property for presets is not mandatory. So you can use it. But you don't have to. I like this discussion for real, so let's keep it up. It's just not really related to this PR. I just made presets channel in slack. Or we can keep it going here as well.

Nice catches from your side, thats why we need your help to review defaults.txt files!
As you can see there is no problem to figure out which parameter is related to TUNE and which one is not.
But there is a problem with my attentiveness, thats why you see `motor_poles`, `gyro_calib_duration`, and other nonsense here.

`set dshot_bidir = OFF`
this one is off only because it is OFF by default. Every tune preset will be free to change that.

haslinghuis:
Yes agree with @etracer65 here, discussion is not resolved. As much as I like some form of this to go forward, this needs resolve first. This feature won't have to use older firmware versions, just start from 4.3.

mikeller:
@ctzsnooze: In general in Betaflight repositories (except for `unified-targets`, which is meant to be self-managed by users), we are waiting for at least 24 hours before mergeing pull requests - this is to allow people in different time zones to participate in discussions.
I would also think that it makes sense to wait with mergeing while there still is an ongoing discussion.
Can I ask you to please adhere to this?

limonspb:
> @ctzsnooze: In general in Betaflight repositories (except for `unified-targets`, which is meant to be self-managed by users), we are waiting for at least 24 hours before mergeing pull requests - this is to allow people in different time zones to participate in discussions.
> I would also think that it makes sense to wait with mergeing while there still is an ongoing discussion.
> Can I ask you to please adhere to this?

Make sense about 24 hours.
The discussion tho was not about this pull request.
`# defaults` has been in presets repo for a while. I just added an option to use multiple of them if needed.

We also want to rename them from `# defaults` to `# included_preset`, to avoid confusion.  Including the other preset file is what really happening when using this property.

mikeller:
In general I think that @etracer65 has a valid point when he says that duplicating defaults outside of the firmware introduces some problems - this is why we have a rule that defaults are not allowed in Unified Targets, for example.
I think that the risk of dangerous configurations can be mostly mitigated by having validation rules in the firmware that will reset combinations of settings that are dangerous. But the problem of creating extra work, and extra potential for inconsistencies definitely exists.
One way to work around this that I can see will be to introduce a way in the firmware to reset the settings only for one or a number of parameter groups. Since parameter groups partition the settings along the lines of features / feature groups, this should allow every type of setting (or even every preset snippet) to specify what groups are to be reset. Obviously the changes will then also have to be 'propagated' by running the firmware configuration validation, as these rules often span multiple parameter groups. This will give us a way to achieve what we are trying to do here, but in a way that is independent of the firmware version.
To take this one step further, if we implemented hierarchical parameter naming (e.g. `get <parameter group>.<parameter name>`), we could automatically extract the names of all affected parameter groups from a snippet, and then reset these groups.
But unless we put in the effort to at least implement parameter group based reset to default for 4.3, I think for now adding and maintaining explicit defaults as part of the presets is the only way to get this to a point where it is releasable.

limonspb:
Agree with @mikeller. There are some nice ways to solve the problem, however, the scope of that is really big, lets see if we can achieve that for 4.4. And I agree that so-called `# defaults` snippets or `# included_presets` is the only releasable way for now. Let's make them for 4.3 as a compromise, define allowed TUNE properties, and list them in `defaults_4.3_TUNE.txt`.
Future 4.4 presets will be able to be completely different: good thing we using CLI for all that, should be pretty flexible.

etracer65:
Another problem with including defaults in the presets that I was about to mention before the discussion was interrupted, is the interaction with the settings in the unified target definition. While we don't allow firmware defaults in the target definitions, most other settings are valid. For example there's no reason why a target that's for a specific bind-n-fly quad might not have settings that will conflict with the "defaults" referenced in the preset. The earlier example of `dshot_burst` and even `motor_poles` are examples of this. But it could carry down to much more mundane settings such as PIDs and rates. So there's no way for the "defaults" associated with the preset to possibly be valid for that particular target. But again, using the firmware mechanism to reset to defaults resolves this. So then you have to go down the rabbit hole of possibly having not only version-specific defaults for each preset, but also target-specific versions for each variant. Obviously this is unmanageable. And we can't say "I don't think this will happen" - it probably already has. Take a look at the target definitions. And further there's nothing to stop it from the future. We can't be in a position to say "don't add something" to the target definition because it might conflict with some preset "defaults" now or in the future.

So if we're still going down the (incorrect I believe) path to apply "custom defaults" associated with presets, only way to resolve this I can see is to try and do some kind of 3-way match when applying the preset that tries to reconcile the current firmware settings, the preset "defaults" and the target definition settings. And this isn't likely to be viable. For example, how do you reconcile conflicts between the preset and the target defaults? Which is "correct"? Using the target config might "break" the preset. Using the preset default might make the target unflyable. No way to know programmatically - particularly with the levels of freedom currently available based on what can go into a target config and preset.

etracer65:
My only major objection is in regards to maintaining "shadow" sets of defaults. That need appears to be driven by the belief that the preset must have a "clean slate" to work properly. So as I already proposed, if that's the case then the preset should require defaults and we should use the existing firmware mechanism to enforce that (which also solves the target definition conflicts). Or alternately the requirement that preset should "reset" unaffected settings should be dropped.

If a more elegant solution can be brought forward in the future (like only affecting certain "groups" of settings), then it can be revisited. Personally I don't think it's severe burden on the users to expect that applying a "preset" should require you to reset your settings. It's no different then upgrading to another firmware version.

mikeller:
@etracer65: The problem that I see is that with the only way we have right now to reset settings, **all** settings will be reset, even the ones that were not affected by the earlier application of the preset. So I think unless we come up with a clever way to make it convenient for the user to save / restore the configuration for their craft (or even do this for them before applying a preset), asking users to do a manual backup every time before they apply a preset, just so they don't lose their current configuration including all previous presets will make it a lot less attractive to use presets.
And yes, I think the 'clean slate' capability is needed - in particular in order for users to be able to 'remove' a preset that they applied, if they don't like it or it does not give them what they want.

etracer65:
As I previously suggested:

> I would suggest that the preset have a flag that determines when a "reset to defaults" is required. Presets that are vulnerable to the user's possibly incompatible settings like extensive tuning should have this set to "yes". Presets that maybe only adjust rates or are less inclusive might not need to make this a requirement. When a user tries to apply a preset that requires it then they can be warned and asked if they want to continue.

mikeller:
@etracer65: This does nothing to make it easy / convenient for the user to backup / restore their settings before / after the reset to defaults is done, and this does nothing to make it possible for them to 'remove' a preset.
Users are not likely to want to use a preset if it means they will lose all of their settings by doing so, and if there is no way to 'remove' a preset if they don't like it.
In this case we'll be offering something that doesn't have much of an advantage over asking users to copy / paste snippets into the CLI, I am not sure if this will be worth the effort.

etracer65:
> This does nothing to make it easy / convenient for the user to backup / restore their settings before / after the reset to defaults is done, and this does nothing to make it possible for them to 'remove' a preset.
> Users are not likely to want to use a preset if it means they will lose all of their settings by doing so, and if there is no way to 'remove' a preset if they don't like it.
> In this case we'll be offering something that doesn't have much of an advantage over asking users to copy / paste snippets into the CLI, I am not sure if this will be worth the effort.

Likewise we have no mechanism to save/restore settings before a firmware update either. So I don't think introducing an unsupportable shadow defaults mechanism is more warranted here. Also there is no mechanism to "remove" a preset in any of this if the user decides they don't like it. As designed their only option would be to hopefully have saved a `diff` before applying the preset, otherwise reset to defaults. The point being that if a preset is comprehensive enough that the author considers it necessary to ensure the user has default settings applied then I don't see the burden of requiring default settings. We have no way to ensure that the "defaults" contained in the preset will actually include all of the things that the user might have actually changed that could negatively affect the performance of the preset. Further it'll be a maintenance nightmare to keep it up to date, police this compliance with "user submitted" presets, etc. We make the same stipulations for version upgrades - start with defaults. And as proposed, if the preset is minor in nature (like rates only) then the defaults application is not needed.

limonspb:
Applying, replacing, removing TUNE preset has to be easy for the user. Otherwise, I do not see a point of it at all. He should be able to quickly try different TUNEs and decide what works the best for him.

@etracer65 we have to define the list of what should be considered as "TUNE" variables and allow only these variables for TUNE category presets. This list either should be defined in the firmware (a better way) or in the preset defaults_tune.txt file.

Also this list must be pretty strict and not that long. Stuff like `dshot_burst` or `motor_poles` aren't allowed.

If the preset author wants to put some CLI commands outside of this list, he has to have a good reason for that. Then he marks preset as MIXED and adds a warning/explanation that there is no easy way to go back unless the user saves diff.

limonspb:
> > Applying, replacing, removing TUNE preset has to be easy for the user. Otherwise, I do not see a point of it at all. He should be able to quickly try different TUNEs and decide what works the best for him.
> 
> Okay, a noble goal but incomplete. What if the user decides that he doesn't like the preset TUNE he tried? How does he "remove" that tune and go back to what he had before? Maybe back to the original settings that shipped with their bind-n-fly? The point is that he can't - unless he happened to save a `diff` before applying the first preset, resets everything to defaults, then reapplies his saved `diff`. So by making it "easy" for users to try out different TUNEs you actually lead them down a path of unknowingly losing their settings.

Yes, if they had some custom tuning, no easy way for them to go back to that after playing with TUNE presets. Unless we make a feature "save/load current tune" button in the future. Which, again, should not be that difficult if we have a defined list of TUNE variables in CLI.

etracer65:
Frankly, the only way I see this entire concept of being viable is with a complete revamp of the settings structure. First, better compartmentalization of the different "classes" of settings. Not sure if this is really completely achievable as there will always be overlap between a setting being part of a "tune" or "preference". Next we need to implement a layered settings system with each subsequent layer able to selectively override the lower levels. The layers from bottom to top would be: firmware defaults, target settings, user preference settings, finally an optional preset. The runtime settings would be the consolidated view with upper layers overriding any defined settings. The bottom layer (firmware) would be complete with every setting. Each subsequent layer is "sparse" with only the changes needed. Then if the user wants to apply a preset then it has no persistent effect on the user's previous settings, target, or firmware defaults. It just overrides at runtime. Want to try another preset then no problem. Remove the preset and you're back to the user's settings. Reset to "defaults" and remove the user's setting layer and you're back to target defaults. Etc.

But no doubt this is a huge change. While we already effectively have the first 3 layers they are not isolated. And settings take flash space and really need to be in their own block so adding another for presets is problematic. Not to mention the challenges in how the user would interact with settings in the Configurator. Like would they not be able to change any of the active settings that are supplied by the preset? If they were then how would we record that - yet another layer?

limonspb:
Yes, a huge challenge with the different _layers_ of presets. But it probably needed for a complete and nice solution.
What we propose for now, i believe with @mikeller and @ctzsnooze is to make some presets that will work for the coming 4.3 (hopefully coming soon). We mark these presets files as "4.3" and they will become automatically obsolete when we release 4.4+. By that time in far 2022 or 2023 if we have a better way to reset/restore/reapply presets - we will add it in CLI.

And the good thing (hopefully) with CLI-based presets is that in the future we don't need to change the configurator code much, but just change/add CLI.txt files with new commands like `restore tune defaults`, `restore rates defaults` - something like that.

And for now, we can help users to play with tune presets, help with applying VTX tables and famous racers rates and some other fun stuff.
And yes at the cost of Ctz time creating and maintaining 4.3_tune_defaults.txt
And at the cost of frustrated users that did apply a TUNE presets, but they want the go back to what they had and forgot to save diff.

mikeller:
@limonspb: Can you have a look if https://github.com/betaflight/betaflight/pull/10755 can be used to mitigate the problems around resetting the configuration?

haslinghuis:
Funny like I proposed 
Screenshot from 2021-05-23 02-20-58

limonspb:
> @limonspb: Can you have a look if [betaflight/betaflight#10755](https://github.com/betaflight/betaflight/pull/10755) can be used to mitigate the problems around resetting the configuration?

YES!!!

mikeller:
I'll get betaflight/betaflight#10755 merged then. :smiley:

PR #12, opened 2021-05-19
Allow regions for CLI commands in presets

limonspb:
Regions - **optional** syntax to divide a CLI preset file into named parts.
Allows configurator user to select what exactly he wants to apply.

Example in the configurator:
image

Regions are optional directives. They can be used to mark a group of CLI commands and give a name for this group. They are similar to C# preprocessor directive `#region`.
When a CLI preset contains regions, a Configurator user will see a list of available regions by their names. This list will appear in the preset dialog once a specific preset is selected. User will be able to select desired regions to apply.

An example of using regions is a BNF quad preset. Manufacturer could add different regions for different radio protocols such as SBUS, Crossfire, Ghost, etc. In this case, a user will be able to select his radio protocol when he applies this preset together with different rc_smoothing settings.

A region starts with `# region begin <region name>` directive. It can be extended with `(checked)` or `(unchecked)` to specify whether this region should be selected by default or not.
Every region should be closed with `# region end`. Between `# region begin <region name>` and `# region end` there is a list of CLI commands included in the region.
Nested regions are not allowed yet.
Regions could be split into multiple parts although it is not recommended for readability.

If you `# include` a preset with the regions inside, these regions will not appear in the final preset.
If you still want to include them, you have to add a dummy region declarations like:
```
# region begin <region name>
# region end
```

Using `# include` is allowed within the region and works as expected:
```
# region begin <region name>
# include file/to/include.txt
# region end
```

syntax example:

```
# Title: BMSThomas racing/freestyle rates
# FirmwareVersion: 4.2
# FirmwareVersion: 4.3
# Category: RATES
# Official: false
# Keywords: racing
# Author: Thomas "BMSThomas" Bitmatta
# Description: Racing and freestyle rates by BMSThomas - MultiGP International Open 2019 champion
# Description: ACTUAL rates with 720 degrees/sec max
# Description: Note: this preset slightly changes throttle curve as well
# Defaults: presets/4.3/rates/defaults_4.3_rates.txt

# region begin (unchecked) Throttle curve
    set thr_mid = 45
    set thr_expo = 30
# region end

# region begin (checked) Rates Yaw Pitch Roll
    set rates_type = ACTUAL
    set roll_rc_rate = 19
    set pitch_rc_rate = 19
    set yaw_rc_rate = 19
    set roll_expo = 56
    set pitch_expo = 56
    set yaw_expo = 56
    set roll_srate = 72
    set pitch_srate = 72
    set yaw_srate = 72
# region end

# region begin Throttle limit
# region end Throttle limit
```

indexer script verifies:
1. All regions must be closed with `# region end`
2. Nested regions are not allowed
3. Region name can't be empty


index_hash.txt
```diff
@@ -1 +1 @@
-46f4dcae39877158b31cc736c8dd32e11ef521775d386e13ab8523080cab05ca
\ No newline at end of file
+3cabb4c1087b654754313ec0113e05b95ed6cb211a4d6e15843d8a38c360ff5f
```
mikeller, line 1: Do we still need the hash? I thought we'd switched to using the git hash to check if we need to reload the file?

mikeller (approved):

index_hash.txt
limonspb: @blckmn proposed to use https://api.betaflight.com/presets/** that will be automatically updating with every change in the git. IN this case I guess we still need index_hash.txt. But I am not using it yet and maybe won't at the begining.

PR #13, opened 2021-05-20
Renaming # defaults to # include

limonspb:
Seems that `# defaults` metaproperty for presets is confusing (https://github.com/betaflight/firmware-presets/pull/11)
I think renaming it to `# include` can help a bit.

So if `# included_preset: path/to/other/file.txt` is used
then all the commands from `path/to/other/file.txt` will be included
Similar to # include in C/C++

Multiple files can be included:
```
# include: path/to/other/file1.txt
<cli command 1>
<cli command 2>
# include: path/to/other/file2.txt
# include: path/to/other/fileN.txt
<cli command 3>
<cli command 4>
```

Indexer script verifies that every `# include:` has a valid and existing file specified.
Recursion is not supported yet in the first approach.


mikeller:
Should we make this generic and call it `include`? I think for most people involved in this this will make it self-explanatory.

Also, not sure about the 'except that included preset will be added at the beginning' - is there any drawback to having this splice the `include` in where it is found in the file - naturally defaults would be listed at the beginning anyway, but this might also be used to declare other composite constructs.

limonspb:
@mikeller i agree can make it just `# include`. The problem is that now it won't go more than 1 step deep in the configurator and it can't detect cyclic references. But that's just some extra coding, probably not that much.

ctzsnooze:
Just `# include` would, indeed, be nice, but if it causes a lot of problems, `# include_preset` would be fine by me also.

mikeller:
@limonspb: I think we can live with a non-recursive implementation of this for now if it gives us what we need for defaults. Recursion can be added later. And without recursion I don't think the aspect of circular references is much of a problem - detecting / preventing files from including themselves can be done manually in reviews.

limonspb:
ok, renamed to `# include`. Changed the readme description.
Let's merge if looks good.
The configurator part is in progress.

ctzsnooze (approved):

ctzsnooze (approved):

PR #14, opened 2021-05-27
some rc smoothing settings

ctzsnooze:
Some possible RC smoothing presets


McGiverGim:
Two files with name ending in "copy"... it seems strange to me. Is that intended?

limonspb:
I would put all the RC_SMOOTHING 4.3 presets under 4.3/RC_SMOOTHING/ folder for a little better organization like:
<BF version>/<category>/<preset_name.txt>

The name "HD freestyle_rc_smoothing.txt" missing the underscore I guess.

I also think that because you don't have automation for indexing yet, we should run it for every pull request manually and add the result to pull request. Otherwise making separate pull requests with index.json changes would be ugly.

presets/TestSamples/4.3/default_rc_smoothing.txt
```diff
@@ -0,0 +1,17 @@
...
+
+set rc_smoothing = ON
+set feedforward_jitter_reduction = 10
+set feedforward_transition = 0
+set rc_smoothing_auto_factor = 30
+set rc_smoothing_auto_factor_throttle = 30
+set rc_smoothing_setpoint_cutoff = 0
+set rc_smoothing_feedforward_cutoff = 0
+set rc_smoothing_throttle_cutoff = 0
+set rc_smoothing_debug_axis = ROLL
```
limonspb, line 17: `set rc_smoothing_debug_axis = ROLL`
do we need this for user presets?

presets/TestSamples/4.3/race_rc_smoothing.txt
```diff
@@ -0,0 +1,19 @@
...
+# Description: RC smoothing settings for racing - assumes a good RC link
+# Description: Slightly increases stick responsiveness over defaults
+# Description: Note: defaults are close to ideal for racing
+# Description: Warning: sends some noise to the motors at link frequency during fast stick movements - may make motors warmer than usual
+
+set rc_smoothing = ON
+set feedforward_jitter_reduction = 5
+set feedforward_transition = 0
+set rc_smoothing_auto_factor = 22
+set rc_smoothing_auto_factor_throttle = 20
```
limonspb, line 16: Let's `# include` defaults file for each preset at the beginning?

For example `rc_smoothing_auto_factor_throttle = 20` is used for race_rc_smoothing.txt
However "cinematic_rc_smoothing.txt" does not set this value at all.

It means that if someone applies "race" and then he wants to try "cinematic", he will have "rc_smoothing_auto_factor_throttle = 20" from the race tune left.

limonspb (changes requested):

presets/TestSamples/4.3/race_rc_smoothing.txt
ctzsnooze: I specifically only overwrote the files that needed overwriting.
The cinematic rc smoothing one, once a manual value is set, the user's old auto smoothing value can stay, since it doesn't matter what the auto value is once you use a manual setting..  That's why I did it that way.  
I could set the auto value to a cinematic value but again specifically didn't want to do that.

presets/TestSamples/4.3/default_rc_smoothing.txt
ctzsnooze: its the default value, so if we reset to defaults, then yes

presets/TestSamples/4.3/race_rc_smoothing.txt
limonspb: @ctzsnooze it's probably a long conversation, but i think that presets are in general for the people who don't mess with individual parameters. They want to try "race_rc_smoothing", then "cinematic" and see what works best. They don't need to think what smoothing parameters will be left. These people don't need any leftovers between applying/trying these presets.

presets/TestSamples/4.3/rc_smoothing/race_rc_smoothing.txt
```diff
@@ -0,0 +1,19 @@
+# Title: Race RC smoothing
+# FirmwareVersion: 4.3
+# Category: RC_SMOOTHING
+# Official: true
+# Keywords: aggressive, rc, smoothing, race
+# Author: Chris Thompson ctzsnooze
+# Description: RC smoothing settings for racing - assumes a good RC link
```
limonspb, line 7: Should we specify what is "good Rc link"? What's the minimum update rate required. Or should we make race presets for different links/update rates?

limonspb (commented):
Now when we moved the files inside the `rc_smoothing` folder, I guess, we don't need "rc_smoothing" as a part of the file name?

presets/TestSamples/4.3/race_rc_smoothing.txt
ctzsnooze: They can do exactly that.  I checked that on each occurrence, what the person writes is what they get.
We don't need to change values that a user may have carefully configured previously, unless we must change them for the preset to work.
If they want to revert to defaults they can run the default first, then the preset.
But I don't want to force them to do so every time.

presets/TestSamples/4.3/rc_smoothing/race_rc_smoothing.txt
ctzsnooze: Yes is difficult.
Maybe "CRSFshot or better"?

limonspb: @ctzsnooze  at least something that explains what is "good". Because for some "good RC link" is the one that goes far (crossfire 50hz), not the one with fast update rate and minimum jitter :)

ctzsnooze: This is difficult because is FrSky a good link?  I think so. 
Maybe I mean a fixed rate link.
I think it might be better to just remove `- assumes a good RC link` :-)

ctzsnooze:
Updated the content and renamed the files.
Will need some kind of identifier in addition to the type of flying, so used my nickname.
If these are all official betaflight presets, then the name structure could be `cinematic_betaflight43.txt` for example.

sugaarK:
only thing you might consider @ctzsnooze  is adding the recommendation in the description that the user updates their radio equipment and checks the ADC is off as if the smoothing presets are not used with the radio link  presets then the use might not become aware of this

PR #15, opened 2021-05-28
add FrSky_sbus preset to rc_link folder

ctzsnooze:
First draft of what I think are the kind of things that should go into an rx link preset.
Example provided is for FrSkySbus.
I anticipate we need them for Crossfire, CRSFshot, ELRS, etc.
The description fields can provide useful information about how to configure the sub-settings.


limonspb:
looks good to me if talking about structure, metaproperties etc.
I would run an indexer script also to add this preset into the index.json

And maybe ask to review someone who knows about these cli commands? Because I am not qualified to review that part :)

limonspb (approved):

PR #20, opened 2021-05-28
Added indexer/check.js to run verifications without creating index.json

limonspb:
usage:
`node indexer/indexer.js` - runs verifications for presets, creates index.json
`node indexer/indexer.js nosave` - just runs verifications
`node indexer/check.js` - just runs verifications, same as `node indexer/indexer.js nosave`

exit code 0 - no errors

### DIFF
--- a/indexer/check.js
+++ b/indexer/check.js
@@ -1,0 +1,6 @@
+'use strict';
+
+const child = require('child_process').fork('indexer/indexer.js', ["nosave"]);
+child.on('exit', (code) => {
+    process.exitCode = code;
+});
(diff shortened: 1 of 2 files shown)


ctzsnooze (approved):
`node indexer/check.js ` is important since it provides a means for a PR contributor to run a validation check on their syntax, but not generate index files.

mikeller:
@limonspb: Can I recommend that you add the execution of these various tasks as `scripts` to `package.json` so that they can be executed from yarn? This will make it possible to run them in a controlled environment, instead of using the first node binary found in the path.

PR #24, opened 2021-10-06
karate race preset tune

ctzsnooze:
This preset is for a 6S 5inch race tune with 4.3 slider support.

It has been used successfully by many racers on a number of race frames.

A stock tune which tolerates fairly beat up builds is the default, and an optional alternate 'spicy' tune for cleaner builds is also provided.


sugaarK:
delete this @ctzsnooze ?

PR #28, opened 2021-10-08
Fixed link error on readme.md

DusKing1:
The current [xx](xxx) way will link to /blob/master/xxx, which does not exist and make no sense.

This PR just fix this, and modify one title to match up


DusKing1:
@ctzsnooze I think now I fix all of them, but do review in case I missed something.

PR #29, opened 2021-10-08
Defaults Presets for 4.3

ctzsnooze:
default presets  for 4.3

 early draft


ctzsnooze:
Question - these initial drafts use regions to provide control over what is being reset.

However if a default preset with regions is included in some other preset, then those options may not get applied.

So... should we **not** use regions in defaults presets?

PR #30, opened 2021-10-08
TBS link presets

sugaarK:
Individual  presets for each TBS radio link and frequency  modes. The crossfire dynamic and 50hz locked may require some proper testing


sugaarK:
I've updated the 50hz CF (which was the one that needed consideration) to match @ctzsnooze's elrs 50hz preset so this is good to go now @limonspb

sugaarK:
done @limonspb

PR #38, opened 2021-10-09
Readme update: Support FILTERS and OTHER categories, multiple other edits

ctzsnooze:
Extensive update to readme

- support for `FILTERS` and `OTHER` categories
- remove `MIXED`
- add notes about `DIscussion` and `Exclude` properties
- multiple clarifications and rearrangements

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ A simple way to configure your Betaflight Flight Controller Firmware settings.
 
 ## Introduction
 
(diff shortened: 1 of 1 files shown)


limonspb:
Thanks a bunch!
Could you please add information on how to use **nosearch**:
it should be `# nosearch: true` or `# nosearch: false`
if there is no # nosearch then by default it is `# nosearch: false`

PR #47, opened 2021-10-11
Add BMI270 filter settings

DusKing1:
Please review and give me feedback/change requests.

### DIFF
--- a/presets/4.3/filters/BMI270.txt
+++ b/presets/4.3/filters/BMI270.txt
@@ -11,14 +11,14 @@
 set gyro_lowpass2_type = PT3
 set gyro_lowpass2_hz = 150
 
-# Region begin (unchecked) BMI270 Clean Build
+# option begin (unchecked) BMI270 Clean build without RPM filters
 set gyro_lowpass_hz = 0
 set gyro_lowpass2_type = PT3
 set gyro_lowpass2_hz = 135
 set dyn_lpf_gyro_min_hz = 0
-# Region end
+# option end
 
-# Region begin (unchecked) BMI270 Clean build with RPM filters
+# option begin (unchecked) BMI270 Clean build with RPM filters
 set gyro_lowpass_hz = 0
 set gyro_lowpass2_type = PT2
 set gyro_lowpass2_hz = 135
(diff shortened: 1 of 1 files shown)


DusKing1:
@sugaarK @limonspb pushed new commits so fixes your suggestion.

@limonspb yes. The BMI270 filter settings I submitted can be applied from 2'' all the way up to X-Class and X8 cinematic class. It's quite safe to just apply them on almost all the builds.

> So let's say i build a 5 inch racer with BMI270, I apply Karate tune from Zak, and then i need to apply this preset on top of that?

That's hard to just say yes or no, since AFAIK the Karate tune which is a pretty aggressive tune is mainly based on MPU6000/ICM20602 whose sampling rate is always like 8k or 4k, but the BMI270 ODR is 3.2k, and there's a cutoff frequency loss in our algorithm. So it's is not very wise that just apply these settings on the different gyroscopes, especially from a 4k/8k ODR to a 3.2K ODR.

> And then if i build X class, use X class preset and then i still go here and apply the same filter settings?

The key thing of filter settings for BMI270 is that it needs a PT2 or PT3 filter on lowpass2 to give a strong attenuation, so yes you can do that.

ctzsnooze:
Please rebase to use `option` in place of `region`.

DusKing1:
So here are our tests.
[BMI270 test series.zip](https://github.com/betaflight/firmware-presets/files/7358616/BMI270.test.series.zip)

The default filter settings on BMI270 is clearly not that good enough. As we accept when the PSD belows the -40db is "low enough", on the BMI270 PSD spectrum we can still see some leaks.
image

As the solution suggests, we can simply apply a PT3 on gyro lowpass 2 so those noises disappear.
image

And if we look for less delay on clean build, we can disable the whole lowpass 1 filter and lower the cutoff frequency on lowpass 2. 135hz PT3 looks like this:
image

Here we enable the RPM filters and change the lowpass filter type to PT2 so the delay can be lower further.
image

And at this stage, it's quite safe to increase the cutoff frequency of Dterm so can increase the whole system's bandwidth to gain more performance on PID controller. Jump from the red to the blue.
image

All plots are plotted by Blackbox Mate, can be visited from https://bbm.pitronic.top.

DusKing1:
Could not be done since current master will cause h743 lock up after power up, as I said before.

DusKing1:
AFAIK currently no fix, and I have no time keep up with all the changes in firmware repo.

SteveCEvans:
@DusKing1 Are the H743 issues you mention described anywhere? Both the IFLIGHT_H743_AIO and MATEKH743 are working perfectly for me.

DusKing1:
@SteveCEvans there are, spread in DMs and threads, lemme collect them in one issue now. See details in https://github.com/betaflight/betaflight/issues/11035

PR #50, opened 2021-10-12
RC_LINK : update frsky with more options for flight style

ctzsnooze:
Includes race, HD freestyle, and cinematic alternatives to general purpose defaults.
Includes an option for people who run more than 8 channel modes, which always sends duplicate data every second packet.  This is not recommended for either racing or freestyle since it is very slow.


ctzsnooze:
The new settings for SBus bound in 16-channel mode - the `Bound with >8 channels` option, 2 point averaging, 35 of smoothing, set give decent results, considering that packets are sent in duplicates every 22ms :-)

Screen Shot 2021-10-12 at 14 28 26

For comparison, similar stick inputs bound in recommended 8-channel mode with the default settings (no averaging and smoothing of 25) give sharper results with more definition

Screen Shot 2021-10-12 at 14 28 48
.

ctzsnooze:
HD Freestyle and cinematic will be very smooth under all conditions, including slower stick movements.

This is a slower move in HD Freestyle:

Screen Shot 2021-10-12 at 14 39 29

Default and Race options will carry some jitter for slow moves, but will be sharper and quicker for fast inputs.  This is a slower move in Race mode, with some shaking (mostly my fingers) getting through the smoothing.

PR #51, opened 2021-10-12
Karate filter array

sugaarK:
this is the karate filter array. It has had extensive testing over the few months on a wide range of machines both race and freestyle based 5" machines.. Vs 4.2 run with no gyro LPF only rpm and the FFT notch this produces around half the gyro filter phase delay using this setup...  While over personally run it on every thing from my betaFPV65 to my 6" LoRanger 99% of other user testing has been on 5" quads


sugaarK:
sorry @limonspb  I sorted this out its correct now

sugaarK:
also lets go.. me and CT have agreed to revisit this with more testing when time permits @limonspb

PR #52, opened 2021-10-12
Add disclaimer and sample text for VTx Preset

ctzsnooze:
For discussion - adding a disclaimer and sample text for the VTx Presets.


sugaarK:
Yeah @limonspb. I spoke to @ctzsnooze about it he’ll fix it up when gets time

ctzsnooze:
Updated the text, @sugaarK please check that it seems OK now.  There may be some other changes required if we change the `# disclaimer` tag structure.

PR #74, opened 2021-10-29
SupaflyFpv Freestyle EasyTune

SupaflyFPV:
SupaflyFPV Freestyle EasyTune

This preset should be applicable to a wide variety of of freetyle builds, and easily tuned using the Master Multiplier Slider.

I think the slider positions correlate quite well with the main Betaflight default when considering the needs of freestyle, and I have taken a policy of leaving any exotic ideas for parameters out and settling for Betaflight defaults and philosophies rather than splitting hairs.

Those 'needs' of freestyle are no great surprise - more PIDs on the pitch axis due to typical freestyle frame design and HD camera...slightly higher D gain for propwash...some of my 'subjective' preferences such as moderate I gains, FF Jitter at 12, Moderate FF, med-heavy static Dterm filtering, reduced Gyro filters to cover the main noise issues robustly and not worry about that which is not there.

RPM option selected as default, if deselected, Non-RPM setup saved. Both setups tested and working well with good performance / robust balance.

Preset Options: 4s and 6s 5” , and a 6/7" Option. They are all made from the same fundamental slider tune.

Can post pics/logs and video to demonstrate if necessary. These numbers have been tested by me and my pilots over the last couple of years and should be about right. A tad on the low side to be safe, easily tweaked with the sliders...

### DIFF
--- a/presets/4.3/tune/SupaflyFPV_Freestyle_5_Inch_EasyTune.txt
+++ b/presets/4.3/tune/SupaflyFPV_Freestyle_5_Inch_EasyTune.txt
@@ -1,0 +1,185 @@
+#$ TITLE: SupaflyFPV Freestyle 5 Inch EasyTune
+#$ FIRMWARE_VERSION: 4.3
+#$ CATEGORY: TUNE
+#$ STATUS: OFFICIAL
+#$ KEYWORDS: Supafly, SupaflyFPV, 4S, 5s, 6S, Freestyle, 5 inch, 5"
+#$ AUTHOR: SupaflyFPV
+#$ DESCRIPTION: Supa-Easy Freestyle Preset - Easily tune freestyle builds using only the Master Multiplier Slider
+#$ DESCRIPTION: Select the option for your build with or without HD camera (eg GoPro) as the base tune
+#$ DESCRIPTION: Half to one click up or down Master Slider to tweak the tune - in OSD or Configurator
+#$ DESCRIPTION: Quad feels loose, click Master up. Quad shakes/oscillates click Master down
+#$ DESCRIPTION: RPM Filtering for best performance (default) - or de-select for Non RPM Setup (check esc compatibility)
+#$ DESCRIPTION: Enjoy!
+#$ DISCUSSION: https://github.com/betaflight/firmware-presets/pull/41
+#$ INCLUDE_WARNING: misc/warnings/en/rpm_filters.txt
+#$ INCLUDE: presets/4.3/tune/defaults.txt
+#$ INCLUDE: presets/4.3/filters/defaults.txt
+
+# -- Non-RPM Filtering --
+
+set dshot_bidir = OFF
+set simplified_gyro_filter = OFF
+set gyro_lpf2_static_hz = 1000
+set gyro_lpf1_dyn_min_hz = 250
+set gyro_lpf1_dyn_max_hz = 600
+set dyn_notch_count = 3
+set dyn_notch_q = 300
+set dyn_notch_min_hz = 100
+set dyn_notch_max_hz = 650
+
+# -- Dterm filtering --
+
+set simplified_dterm_filter = OFF
+set dterm_lpf1_dyn_min_hz = 0
+set dterm_lpf1_dyn_max_hz = 150
+set dterm_lpf1_dyn_expo = 0
+set dterm_lpf1_static_hz = 90
+
+# -- Antigravity --
+
+set anti_gravity_gain = 4000
+
+# -- Feedforward jitter reduction --
+
+set feedforward_jitter_reduction = 12
+
+#$ OPTION BEGIN (CHECKED): RPM filtering
+
+set dshot_bidir = ON
+set simplified_gyro_filter = OFF
+set rpm_filter_harmonics = 1
+set gyro_lpf2_static_hz = 1000
+set gyro_lpf1_dyn_min_hz = 250
+set gyro_lpf1_dyn_max_hz = 600
+set gyro_lpf1_dyn_expo = 8
+set dyn_notch_count = 2
+set dyn_notch_q = 400
+set dyn_notch_min_hz = 100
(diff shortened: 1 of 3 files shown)


sugaarK:
@SupaflyFPV  what do you think about breaking it up into a 5 and 6/7” tune? We could do with a few more tunes in the presets right now

SupaflyFPV:
> @SupaflyFPV what do you think about breaking it up into a 5 and 6/7” tune? We could do with a few more tunes in the presets right now

yes no probs - I'll split it...

Also adding HD camera and Non HD camera options due to effect on pitch...

limonspb (commented):
No complaints from my side about syntax.
Why add PID values together with sliders tho?

sugaarK (commented):
So the rpm vs non rpm tunes.. I think you could use a fixed 500hz lpf instead of the cascaded pt1s which would be quite a bit less delay. And in your options for different kv. The notes on motor KV won’t be seen unless they are in the txt after the option begin line

limonspb:
Yes I think the KV should be mentioned in the description.

SupaflyFPV:
> So the rpm vs non rpm tunes.. I think you could use a fixed 500hz lpf instead of the cascaded pt1s which would be quite a bit less delay. And in your options for different kv. The notes on motor KV won’t be seen unless they are in the txt after the option begin line

ok cool I'll reconsider the motor kv notes...

The low passes...I have the 1000hz on the LPF2 because many people are still running 4k pid loop so that is there as according to CTZSnooze it is far better to do this job as a down-sampling filter than the otherwise two-point avergae. Maybe it will help a little with electrical noise etc...
The dynamic LPF2 is over those freq because I can't be sure how noisy someones frame is, so my thinking is to put it there where the attenuation starts in the motor band frequencies but on throttle it should be raised enough to not significantly contribute to propwash...I tried it and it seems fine with regard to performance. On my builds I do what you say a static one but when its at 5-600Hz it's not contributing much to noise attenuation so the question is 'how' high performance do you want a general tune applicable for any frame...

SupaflyFPV:
> Yes I think the KV ranges should be mentioned in the description. It also would probably make sense to just make different presets for 4S and 6S and even title them accordingly: SupaflyFPV Freestyle EasyTune 6S 5'' SupaflyFPV Freestyle EasyTune 4S 5'' The more targeted preset the easier for the user :) Also, it's easier for the user to find a preset without the necessity of dealing with options.
> 
> If your presets have a lot in common and you do not want to copy-paste a lot of text, you can make a `#$ HIDDEN: true` preset, that you just `#$ INCLUDE` in all your presets. But it's optional. Keep in mind, #$ INCLUDE **does** include only CLI commands and other #$ INCLUDEs, not the attributes like TITLE, DESCRIPTION etc.
> 
> Personally I do not like the idea that the user HAS to select options, otherwise he is in trouble applying wrong settings :) I think `#$ OPTIONS` in general are not really for making generic presets that fit different quads, but they are for small preset modifications, like final optional tweaks: rpm/no rpm, extra Spicy tune for brave people, extra settings for a particular prop, gopro on/off, top or bottom mounted battery etc. Like if a user forgers to select options - a preset should still work ok. Where as if we make #$ OPTIONS: tinywhoop and 5 inch, if the user forgets to select - he is in trouble :)
> 
> **But this all is pretty personal,** if the author thinks it's better for the end-user to have a preset with options between 6S and 4S and other options - it is up to him. Just need to make sure he mentioned **the necessity** of choosing options in the description.

Ok cool...

So what I started to realise with using the sliders and this preset is that once you have the basic ratios/setup most tuning is there and you simply increase/decrease the Master Slider...so for that reason I was wrapping up the 4s and 6s which are basically the same but two clicks on the Master Slider. So it seemed quite inefficient to separate that into presets. Also - I wanted people to use the Master slider to tweak the tune because it will likely improve half a click up or 1 click. We have two new tuning aids colliding at the same time - sliders and presets....

However - if you think for preset usage it will be better to have separate 4s and 6s presets no problem - I can split them. I did that already for the 5" and seperate 6/7"....I also introduced the complication of whether someone is running an HD camera or not - which adds a significant extra weight on the pitch axis....the problem is that when you add options they start to multiply...

Fundamentally, we could try both approaches and see which one is best in reality...I think its better to start using things rather then try to guess which is the best way first...

SupaflyFPV:
RPM and Non RPM EasyTune PSD

This above is the tune RPM vs Non RPM filtering. Both very similar results.

RPM vs Non RPM EasyTune Spect

This above is the filtered Gyro and PIDsums for RPM and Non RPM

RPM v Non RPM Pre vs Gyro

This above is Pre and Post FIlter Gyro for each. This build I'm using is a bit clean so can farm out the tunes for further testing...

sugaarK:
For sure you just  create a new PR to edit the file

sugaarK (dismissed):
looks good to me @SupaflyFPV thanks for the data as well

limonspb:
@sugaarK  plz don't merge yet. It needs to be squashed and i think @SupaflyFPV wanted some other changes.

SupaflyFPV:
> @sugaarK plz don't merge yet. It needs to be squashed and i think @SupaflyFPV wanted some other changes.

yes I will separate the 6" and 7" Tunes, and then sort the commits...

sugaarK:
@limonspb i wasn’t going to merge it

sugaarK (changes requested):
the titles should include the prop size? easy tune 5 6 and 7"

SupaflyFPV:
> the titles should include the prop size? easy tune 5 6 and 7"

yes thats in the title...just need to squash now...

sugaarK (dismissed):
I think these look good so if @limonspb approves I'll leave it for him to merge

sugaarK:
don't forget to squash the commits @SupaflyFPV

SupaflyFPV:
When merge @sugaarK

sugaarK (approved):
looks good

PR #78, opened 2021-11-04
Added a single option selectable vtx table preset for all Rush SA2.1 VTX's

Directory:
VTX tables for the Rush Tiny tank, tank ult 2, tank ult mini, tank ult plus, tank racing, and tank solo (wish they didnt have such confusing naming) i find rushfpv vtx's extremely frustrating for the amount of money you pay for them because their website (which is literally on their promotional stickers) doesn't exist and the only way i got vtx tables was by emailing support directly. the way i got the support email to begin with was black magic in itself. alright rant over.

the couple rush vtx table presets that where already there where very poorly documented and confusing so i made it all over again. rush's frequencies tables are the same for all their vtxs's and all of them have a 37 and 48ch variant so i turned that into an option. and the only other difference are the power levels between different models so i made those options too. now is the time for me to say that the way the 4.3 presets system does options is disappointing since there's no way to make options required and grouped and single selectable etc. rather than having it be 37ch by default with 48ch an option i would have preferred force the user to pick between a 37 or 48ch option. same for how i did multi model selection. if a user selects multiple models as options only the last one selected will apply. you can see how i worked around this in the preset config to have no repeating lines because i like clean tidy code. i added a disclaimer "do not select multiple models" so users dont try accidentally. i have the tank ultimate's selected by default because i think its the most popular rush vtx model. oh yea all 3 of the tank ultimate's (mini, plus, II) have the same power levels therefore are identical.

please confirm if i properly did the options stuff in the preset.


Directory:
checks are saying the options are nested. did i mess up the syntax or is that just an error?

presets/4.3/vtx/rush_sa2.1_vtx_tables.txt
```diff
@@ -0,0 +1,48 @@
...
+vtxtable bands 5
+vtxtable channels 8
+vtxtable band 1 BAND_A   A CUSTOM  5865 5845 5825 5805 5785 5765 5745 5725
+vtxtable band 2 BAND_B   B CUSTOM  5733 5752 5771 5790 5809 5828 5847 5866
+vtxtable band 3 BAND_E   E CUSTOM  5705 5685 5665 5665 5885 5905 5905 5905
+vtxtable band 4 AIRWAVE  F CUSTOM  5740 5760 5780 5800 5820 5840 5860 5880
+vtxtable band 5 RACEBAND R CUSTOM  5658 5695 5732 5769 5806 5843 5880 5917
+vtxtable powerlevels 4
+
+#$ OPTION BEGIN: (UNCHECKED) 48CH
```
sugaarK, line 24: should be `#$ OPTION BEGIN (UNCHECKED):`

presets/4.3/vtx/rush_sa2.1_vtx_tables.txt
```diff
@@ -0,0 +1,48 @@
...
+vtxtable powervalues 14 23 27 29
+vtxtable powerlabels 25 200 500 800
+# OPTION END
+
+#$ OPTION BEGIN: (UNCHECKED) Tiny Tank
+vtxtable powervalues 14 20 23 25
+vtxtable powerlabels 25 100 200 350
+# OPTION END
+
+#$ OPTION BEGIN: (UNCHECKED) Tank Racing
```
sugaarK, line 40: and here

presets/4.3/vtx/rush_sa2.1_vtx_tables.txt
```diff
@@ -0,0 +1,48 @@
...
+vtxtable band 3 BAND_E   E CUSTOM  5705 5685 5665 5645 5885 5905 5925 5945
+vtxtable band 6 LOWRACE  L CUSTOM  5362 5399 5436 5473 5510 5547 5584 5621
+# OPTION END
+
+#$ OPTION BEGIN: (CHECKED) Tank Ultimate Plus/Mini/II
+vtxtable powervalues 14 23 27 29
+vtxtable powerlabels 25 200 500 800
+# OPTION END
+
+#$ OPTION BEGIN: (UNCHECKED) Tiny Tank
```
sugaarK, line 35: and here

presets/4.3/vtx/rush_sa2.1_vtx_tables.txt
```diff
@@ -0,0 +1,48 @@
...
+vtxtable band 5 RACEBAND R CUSTOM  5658 5695 5732 5769 5806 5843 5880 5917
+vtxtable powerlevels 4
+
+#$ OPTION BEGIN: (UNCHECKED) 48CH
+vtxtable bands 6
+vtxtable band 3 BAND_E   E CUSTOM  5705 5685 5665 5645 5885 5905 5925 5945
+vtxtable band 6 LOWRACE  L CUSTOM  5362 5399 5436 5473 5510 5547 5584 5621
+# OPTION END
+
+#$ OPTION BEGIN: (CHECKED) Tank Ultimate Plus/Mini/II
```
sugaarK, line 30: same here

presets/4.3/vtx/rush_sa2.1_vtx_tables.txt
```diff
@@ -0,0 +1,48 @@
...
+vtxtable powervalues 14 20 23 25
+vtxtable powerlabels 25 100 200 350
+# OPTION END
+
+#$ OPTION BEGIN: (UNCHECKED) Tank Racing
+vtxtable powervalues 14 17 23 27
+vtxtable powerlabels 25 50 200 500
+# OPTION END
+
+#$ OPTION BEGIN: (UNCHECKED) Tank Solo
```
sugaarK, line 45: and here 

so these replace the single rush tables I made?

sugaarK (changes requested):
the correct syntax is 

 `#$ OPTION BEGIN (UNCHECKED):`

presets/4.3/vtx/rush_sa2.1_vtx_tables.txt
Directory: yes they include all the models. all smart audio 2.1. also the docs in the readme are wrong then. i copied the example from that

sugaarK: do you want to fix the readme too ?

Directory:
well the good news is i fixed the syntex issue. the bad news is git did something weird when i tried squashing the fix

presets/4.3/vtx/rush_sa2.1_vtx_tables.txt
Directory: the readme and PR instructions say to only do 1 thing at a time for every pull so ill honor that. maybe another pull.

Directory:
alright excuse the turbulence. git shenanigan's and test suite failures resolved. i also fixed the readme documentation.

sugaarK (approved):
given the readme change is so small I'm ok with you fixing it here 
other wise they look good

limonspb:
this is great!

sugaarK:
@Directory  what are your thoughts on doing this for the TBS vtxs as well ?

Directory:
I don't own any tbs vtx's nor do i know the complicated model naming schemes or have the official tables. Can't test it sorry. Ill just trust whatever's already there is accurate.

sugaarK:
The tables Are from offical literature
Which is more than you can get from rush tank. 😂 the charts we have are good but doing it at a single preset is more efficient

MrAlucardDante:
I updated the preset on my repo (https://github.com/MrAlucardDante/firmware-presets/blob/update_rush_vtx/presets/4.3/vtx/rush_sa2.1_vtx_tables.txt) so it includes No-HAM/EU regulations. I also added option groups.

Let me know what you think about it, and if it's OK for you, I can make a PR.

Angentu:
Hi, dose any one have backup for vtx sologood 2.5w MAX PAx2

PR #84, opened 2021-11-14
CGo rates preset

sugaarK:
Racing rates from Chris Go (Cgo), fastest pilot in NSW Australia and leader of the Quaddiction krew.


presets/4.3/rates/CGo.txt
```diff
@@ -0,0 +1,22 @@
+#$ TITLE: BMSThomas CGo race rates
```
limonspb, line 1: BMSThomas?

limonspb (changes requested):

limonspb (approved):

PR #88, opened 2021-11-23
index.json update

limonspb:
api.betaflight.com not available yet from the Android due to api.betaflight.com "security issues" (need CORS). So we need to keep the Github index.json up-to-date till that gets fixed, hopefully by @blckmn .


butterSnake (approved):
:D

butterSnake (approved):
Lols

index.json
```diff
@@ -92,57 +116,104 @@
...
       "filters",
       "freestyle",
       "freestyle_HD",
+      "freesytle",
       "frsky",
       "hd",
+      "hot",
       "jitter",
       "karate",
+      "karatreBrot",
```
haslinghuis, line 195:
```suggestion
      "karateBrot",
```

index.json
```diff
@@ -92,57 +116,104 @@
...
       "names",
+      "nano",
+      "no rpm",
+      "noisy",
+      "normal",
+      "partial",
       "pid",
+      "pro32",
       "profile",
+      "quaddiction",
```
haslinghuis, line 214: Is this The RCAddict?

haslinghuis:
Instead of putting al in one file growing and growing more and more, would it be possible to have it in separate files?

haslinghuis:
We need to administer `Access-Control-Allow-Origin: URL`

blckmn:
> We need to administer `Access-Control-Allow-Origin: URL`

This is now sorted :)

blckmn (approved):

sugaarK (commented):
@limonspb is this sorted now?

index.json
limonspb: that's in @sugaarK  preset
index.json is generated automatically.

index.json
limonspb: same, index.json is generated automatically.
So that's again a question to @sugaarK  if hr wants to create a PR to fix that.

limonspb:
> Instead of putting al in one file growing and growing more and more, would it be possible to have it in separate files?

It is not all in one. Index.json contains only headers of the presets. The size of this file is roughly 1kb per preset.
By the time we have 1000 presets this file will be only about 1mb.

For 10.9 we should make a new preset source - online SQL database. But for the next few years, we are pretty safe with files.

limonspb:
> @limonspb is this sorted now?

yes don't need to merge this anymore, thanks!

index.json
sugaarK: Name of our local club.. it’s spelt correct

limonspb:
not needed as api.betaflight.com is fixed now

PR #91, opened 2021-11-26
UAV Tech rates

limonspb:
uavTech

### DIFF
--- a/presets/4.3/rates/UAV_tech.txt
+++ b/presets/4.3/rates/UAV_tech.txt
@@ -1,0 +1,23 @@
+#$ TITLE: UAV Tech rates
+#$ FIRMWARE_VERSION: 4.3
+#$ CATEGORY: RATES
+#$ STATUS: OFFICIAL
+#$ KEYWORDS: UAV Tech, youtube, freestyle, HD, cinematic
+#$ AUTHOR: Mark Spatz (UAV Tech)
+#$ DESCRIPTION: 
+#$ DESCRIPTION: Freestyle rates by Mark Spatz (UAV Tech)
+#$ DESCRIPTION: 1000 degrees/sec for roll and pitch, 700 - for yaw. Very smooth around the sticks middle.
+#$ DESCRIPTION: Mark is a Betaflight contributor and tuner who prioritizes data over feelings. His YouTube channel includes some deep dives into tuning and blackbox analyzing.
+#$ DISCUSSION: https://github.com/betaflight/firmware-presets/pull/91
+#$ INCLUDE: presets/4.3/rates/defaults.txt
+
+set rates_type = ACTUAL
+set roll_rc_rate = 1
+set pitch_rc_rate = 1
+set yaw_rc_rate = 1
+set roll_expo = 54
+set pitch_expo = 54
+set yaw_expo = 54
+set roll_srate = 100
+set pitch_srate = 100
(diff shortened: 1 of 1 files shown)


ctzsnooze:
These are plainly not 'race' rates.  They are freestyle and probably also suitable for cinematic or HD filming also.

It would be good to include the keywords `freestyle`, `HD`, `cinematic` I think.

I don't think `youtube` is appropriate as a keyword :-)

limonspb:
> These are plainly not 'race' rates. They are freestyle and probably also suitable for cinematic or HD filming also.
> 
> The description should probably say "Freestyle rates by...."
> 
> It would be good to include the keywords `freestyle`, `HD`, and `cinematic`.
> 
> I don't think `youtube` is appropriate as a keyword :-)

Done! I also think it's appropriate to change them to Official as Mark is a BF contributor and an active "skunk" participant.

ctzsnooze:
We would not include 'advertising' like this, especially in `OFFICIAL` presets:

```
#$ DESCRIPTION: UAV Tech - a YouTube channel about FPV with some deep dives into tuning and blackbox analyzing. Mark is also a Betaflight contributor that prioritizes data over feelings.
```

The description in an `OFFICIAL` preset should be about the flight characteristics of the rates, it's advantages or whatever. It should not be advertising for the youtube channel of the contributor.  

Even in a `COMMUNITY` preset, I think we should be cautious about accepting 'advertising' of that kind.  

Our goal with Betaflight hosted Presets is not to provide an advertising for products or services.  

If we allow this, we probably will get flooded with useless presets intended primarily for advertising, which may well not be a useful addition to the range of presets provided.

ctzsnooze (approved):

presets/4.3/rates/UAV_tech.txt
```diff
@@ -0,0 +1,23 @@
+#$ TITLE: UAV Tech rates
+#$ FIRMWARE_VERSION: 4.3
+#$ CATEGORY: RATES
+#$ STATUS: OFFICIAL
```
sugaarK, line 4: This should be community

presets/4.3/rates/UAV_tech.txt
```diff
@@ -0,0 +1,23 @@
...
+#$ KEYWORDS: UAV Tech, youtube, freestyle, HD, cinematic
```
sugaarK, line 5: I agree with CT here remove the YouTube.

presets/4.3/rates/UAV_tech.txt
```diff
@@ -0,0 +1,23 @@
...
+#$ AUTHOR: Mark Spatz (UAV Tech)
+#$ DESCRIPTION: 
+#$ DESCRIPTION: Freestyle rates by Mark Spatz (UAV Tech)
+#$ DESCRIPTION: 1000 degrees/sec for roll and pitch, 700 - for yaw. Very smooth around the sticks middle.
+#$ DESCRIPTION: Mark is a Betaflight contributor and tuner who prioritizes data over feelings. His YouTube channel includes some deep dives into tuning and blackbox analyzing.
```
sugaarK, line 10: I’m ok with this Mark has been part of the works for a long time and is considered the gateway between BF and YouTube.

sugaarK (changes requested):

PR #107, opened 2021-12-08
Add Diatone Mamba Ultra 1000 VTX tables

NemesisXB:
Diatone Mamba Ultra 1000 VTX tables HAM and NO HAM options.

Link to official VTX table json file: https://cdn.shopify.com/s/files/1/0027/2708/4144/files/Ultra_1000.rar

### DIFF
--- a/presets/4.3/vtx/diatone_mamba_ultra_1000.txt
+++ b/presets/4.3/vtx/diatone_mamba_ultra_1000.txt
@@ -12,21 +12,7 @@
 #$ INCLUDE_DISCLAIMER: misc/disclaimer/en/vtxtable.txt
 #$ DISCUSSION:  https://github.com/betaflight/firmware-presets/pull/107
 
-# reset basic Vtx settings
-set vtx_channel = 0
-set vtx_power = 0
-set vtx_low_power_disarm = OFF
-set vtx_freq = 0
-set vtx_pit_mode_freq = 0
-set vtx_halfduplex = ON
-set vtx_spi_bus = 0
-
-# reset VTx table
-vtxtable bands 0
-vtxtable channels 0
-vtxtable powerlevels 0
-vtxtable powervalues
-vtxtable powerlabels
+#$ INCLUDE: presets/4.3/vtx/defaults_vtx_tables.txt
 
 # Configure common vtxtable settings
 vtxtable bands 5
--- a/presets/4.3/vtx/diatone_mamba_ultra_1000.txt
+++ b/presets/4.3/vtx/diatone_mamba_ultra_1000.txt
@@ -1,0 +1,52 @@
+#$ TITLE: Diatone Mamba Ultra 1000 VTX Tables
+#$ FIRMWARE_VERSION: 4.2
+#$ FIRMWARE_VERSION: 4.3
+#$ CATEGORY: VTX
+#$ STATUS: COMMUNITY
+#$ KEYWORDS:  vtx, vtx table, diatone, mamaba, ultra 1000
+#$ AUTHOR: NemesisXB
+#$ DESCRIPTION: VTX tables for the Diatone Mamba Ultra 1000.
+#$ DESCRIPTION: Select the the HAM option if you have a HAM license or the NO HAM if you do not.
+#$ DESCRIPTION: EU users select the NO HAM option.
+#$ DISCLAIMER:  Only select one option. All previous VTX Table settings will be reset.
+#$ INCLUDE_DISCLAIMER: misc/disclaimer/en/vtxtable.txt
+#$ DISCUSSION:  https://github.com/betaflight/firmware-presets/pull/107
+
+# reset basic Vtx settings
+set vtx_channel = 0
+set vtx_power = 0
+set vtx_low_power_disarm = OFF
+set vtx_freq = 0
+set vtx_pit_mode_freq = 0
+set vtx_halfduplex = ON
+set vtx_spi_bus = 0
+
+# reset VTx table
+vtxtable bands 0
+vtxtable channels 0
+vtxtable powerlevels 0
+vtxtable powervalues
+vtxtable powerlabels
+
+# Configure common vtxtable settings
+vtxtable bands 5
+vtxtable channels 8
+vtxtable powerlevels 5
+vtxtable powervalues 25 100 200 400 600
+vtxtable powerlabels 25 200 400 800 1W
+
+#$ OPTION BEGIN (UNCHECKED): NO HAM
+vtxtable bands 4
+vtxtable band 1 BOSCAM_A A CUSTOM  5865 5845 5825 5805 5785 5765 5745    0
+vtxtable band 2 BOSCAM_B B CUSTOM  5733 5752 5771 5790 5809 5828 5847 5866
+vtxtable band 3 FATSHARK F CUSTOM  5740 5760 5780 5800 5820 5840 5860    0
+vtxtable band 4 RACEBAND R CUSTOM     0    0    0 5769 5806 5843    0    0
+#$ OPTION END
+
+#$ OPTION BEGIN (CHECKED): HAM
+vtxtable band 1 BOSCAM_A A CUSTOM  5865 5845 5825 5805 5785 5765 5745 5725
+vtxtable band 2 BOSCAM_B B CUSTOM  5733 5752 5771 5790 5809 5828 5847 5866
+vtxtable band 3 BOSCAM_E E CUSTOM  5705 5685 5665    0 5885 5905    0    0
+vtxtable band 4 FATSHARK F CUSTOM  5740 5760 5780 5800 5820 5840 5860 5880
+vtxtable band 5 RACEBAND R CUSTOM  5658 5695 5732 5769 5806 5843 5880 5917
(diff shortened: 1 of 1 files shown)


presets/4.3/vtx/diatone_mamba_ultra_1000.txt
```diff
@@ -0,0 +1,52 @@
...
+#$ KEYWORDS:  vtx, vtx table, diatone, mamaba, ultra 1000
+#$ AUTHOR: NemesisXB
+#$ DESCRIPTION: VTX tables for the Diatone Mamba Ultra 1000.
+#$ DESCRIPTION: Select the the HAM option if you have a HAM license or the NO HAM if you do not.
+#$ DESCRIPTION: EU users select the NO HAM option.
+#$ DISCLAIMER:  Only select one option. All previous VTX Table settings will be reset.
+#$ INCLUDE_DISCLAIMER: misc/disclaimer/en/vtxtable.txt
+#$ DISCUSSION:  https://github.com/betaflight/firmware-presets/pull/107
+
+# reset basic Vtx settings
```
sugaarK, line 15: same as other PR please use the default file not manually reseting 

`#$ INCLUDE: presets/4.3/vtx/defaults.txt`

presets/4.3/vtx/diatone_mamba_ultra_1000.txt
```diff
@@ -0,0 +1,52 @@
...
+set vtx_pit_mode_freq = 0
+set vtx_halfduplex = ON
+set vtx_spi_bus = 0
+
+# reset VTx table
+vtxtable bands 0
+vtxtable channels 0
+vtxtable powerlevels 0
+vtxtable powervalues
+vtxtable powerlabels
```
sugaarK, line 29: as above

sugaarK (changes requested):

presets/4.3/vtx/diatone_mamba_ultra_1000.txt
NemesisXB: This not only clears the vtx table but also the vtx stuff:
vtx 0 0 0 0 0 900 900
vtx 1 0 0 0 0 900 900
vtx 2 0 0 0 0 900 900
vtx 3 0 0 0 0 900 900
Which I don't think it should.

sugaarK: why not? the idea is to reset every thing... that's why its a default file

sugaarK: @limonspb  what are your thoughts

NemesisXB: > why not? the idea is to reset every thing... that's why its a default file

I will add it if this holds up the merge. But my thought was that the user probably already has this configured in a way I would not be able to configure on his behalf. If a user for example swaps a vtx out, he would still want the VTX settings configured the same, what could happen is that the new vtx might not have the same number of power levels and the user might have to correct that, but loading the table from the json file would have had the same issue.

In the end using this preset is like using the json file. You just don't have to go and look for it and download it.

NemesisXB: > why not? the idea is to reset every thing... that's why its a default file

Also, as per the readme:
"Ignore some values, or not change values that don't matter to your Preset, at whatever value the user the user has in their configuration, by not mentioning those parameters in the Preset."

limonspb: Hi, @NemesisXB. Thanks for your pull request.
I understand why you do not want to reset 
vtx 0 ... and vtx 1... settings. It makes sense.
But at the same time, why do you want to reset these settings?
```
set vtx_low_power_disarm = OFF
set vtx_pit_mode_freq = 0
set vtx_halfduplex = ON
```

NemesisXB: > Hi, @NemesisXB. Thanks for your pull request. I understand why you do not want to reset vtx 0 ... and vtx 1... settings. It makes sense. But at the same time, why do you want to reset these settings then?
> 
> ```
> set vtx_low_power_disarm = OFF
> set vtx_pit_mode_freq = 0
> set vtx_halfduplex = ON
> set vtx_spi_bus = 0
> ```
> 
> you do not have to change anything yet, let's discuss :)

I'm 100% for not changing these as well. I actually copied everything from the default file except the vtx 0 ... and vtx 1 ... stuff, as I do not really know the other options.

I am not 100% sure about the vtx_halfduplex. This VTX uses IRC tramp which I assume is half duplex (as it has a single data wire) so setting the vtx_halfduplex to on still seems OK. And vtx_spi_bus, I don't know what this is for, I assume vtx controlled via spi? But the first 2 I also think should not be changed by a vtx table preset.

In the end, you probably know better than me what a user might want and not want. And I will update the PR to your requirement. You are after all the guys who put in all the work, and if I'm not mistaken, presets is your baby.

sugaarK: @NemesisXB  ive just created a vtxtable only default. once its merged lets use that..

limonspb: @NemesisXB
Please have a look at this pull request:
https://github.com/betaflight/firmware-presets/pull/109
see if you think that these defaults would be cool for VTX tables.

NemesisXB: > @NemesisXB Please have a look at this pull request: #109 see if you think that these defaults would be cool for VTX tables. Then we merge it and you can use new defaults that resets ONLY vtx tables.

@limonspb, Looks good to me. Will update as soon as it is merged.

sugaarK: @NemesisXB tiz merged

presets/4.3/vtx/diatone_mamba_ultra_1000.txt
```diff
@@ -0,0 +1,38 @@
...
+#$ DISCUSSION:  https://github.com/betaflight/firmware-presets/pull/107
+
+#$ INCLUDE: presets/4.3/vtx/defaults_vtx_tables.txt
+
+# Configure common vtxtable settings
+vtxtable bands 5
+vtxtable channels 8
+vtxtable powerlevels 5
+vtxtable powervalues 25 100 200 400 600
+vtxtable powerlabels 25 200 400 800 1W
```
limonspb, line 22: I am curious why power levels do not match with the power labels.... And i see it is like that in the JSON file too 0_o
Do you have this VTX? Could you check if that's SA 2.1 or SA 2.0?
Configurator 4.3 shows SA version on vtx tab on the right side.

NemesisXB: I have a friend with one, but he is not running 4.3. I know it is Tramp, does it also have version number?

The manual also shows the same values, so it is not just in the json file.
image

limonspb: We can merge this no problem as it matches with the official tutorial.
Maybe it's too much to ask, could you email to Diatone, say that you work on Betaflight Preset and you need to clarify why the power levels don't match with the labels?

For HDZero it was kinda a similar problem we found, and @sugaarK emailed to them and they found a bug in the firmware thanks to us LOL

NemesisXB: > We can merge this no problem as it matches with the official tutorial. Maybe it's too much to ask, could you email to Diatone, say that you work on Betaflight Preset and you need to clarify why the power levels don't match with the labels?
> 
> For HDZero it was kinda a similar problem we found, and @sugaarK emailed to them and they found a bug in the firmware thanks to us LOL

Yeah sure. Can do.

NemesisXB: @limonspb, Diatone replied pretty quickly. but I don't understand it anyway.

This is their response:
The power setting and the value of the image transmission do not necessarily correspond. It depends on the program setting. The image transmission manual is directly marked. Can be seen.

-----------------
"Image transmission power setting value" and image transmission (hardware) in BF, subject to the image transmission (hardware) program

For example: the power in the image transmission program is: 400, 600, 800, 1000
Should be entered in BF: 400, 600, 800, 1000

If you enter other values in BF, the image transmission (400, 600, 800, 1000) will prevail

limonspb: it is hard to understand yeah :)
I guess we should adjust values and labels so they match with the true VTX output power and be done with it.
Could you please also add into title something like Diatone Mamba Ultra 1000 VTX Tables (SA2.0)
And SmartAudio 2.0 in the description?
In case they release a new version of the vtx that is 2.1 - that happens sometimes. At least people will see that this vtx is 2.0 only.
And i guess we are done :)

NemesisXB: @limonspb, why SA2.0? This VTX uses TRAMP.

limonspb (approved):

PR #125, opened 2021-12-23
Fix TUNE and FILTERS defaults to match with latest 4.3 defaults

limonspb:
Well, i hope no more changes after that :)


haslinghuis (approved):

haslinghuis:
You could assign the values with the defaults defined in `fc.js` to make it dynamically update with feature changes.

limonspb:
> You could assign the values with the defaults defined in `fc.js` to make it dynamically update with future changes.

I think in future we need to get rid of hardcoded defaults in configurator code. Only firmware knows what are the defaults and Configurator, LUA, other MSP clients should not worry about that.

sugaarK (approved):

PR #130, opened 2021-12-26
Delete karate_race.txt

sugaarK:
taking the karate race off the presets while we address some issues with it


limonspb:
sorry, my bad

PR #134, opened 2021-12-29
Hdzero fix

sugaarK:
uses defaults_vtx_tables.txt now


limonspb (changes requested):
i think this should not change Karate_race :)

limonspb (approved):

PR #141, opened 2022-01-05
Add bmi270 filter settings

DusKing1:


limonspb (approved):

limonspb (approved):

limonspb (approved):

limonspb:
some test on the old BF4.3 are here:
https://github.com/betaflight/firmware-presets/pull/47

for future, we can redo the test on release 4.3, but for now, i think, this preset could be experimental, and let's see the feedback.

KarateBrot (approved):

ctzsnooze (approved):

KarateBrot (approved):

babang1da:
set gyro_hardware_lpf = OPTION_2 does not work

sugaarK:
> set gyro_hardware_lpf = OPTION_2 does not work
> Allowed values: NORMAL, EXPERIMENTAL

you must be using an older 4.3 build one of the rc builds perhaps... the new modes are defiantly in the 4.3 release

rochford77:
please dont take this the wrong way, but this preset along with https://github.com/betaflight/firmware-presets/pull/182/ should now be marked as 4.2 or removed IMO. 

THis preset was SUPER helpful before we had with osr4/"normal"mode, but now with the lower internal cutoff there are no longer special considerations to take with the bmi 270 compared with other popular gyros like the MPU-6000. This preset simply existing as a 4.3 preset just adds confusion to people flashing. I see people ask at least once a day if they need this preset on various socials, and those who flash without asking end up severely over-filtered, with a worse flying quad.

This still provides value as a 4.2 preset.

sugaarK:
removing these due to constant complaints about them

PR #142, opened 2022-01-08
whoop tune by Justice (Gary Kent)

ctzsnooze:
Official whoop tune based on Gary Kent's carefully tuned 65mm 1S build.

The default values are a little softer than the 'spicy' option that Gary normally flies.  Both are quite aggressive and require a really good strong build.

A 'safer' tune option is provided for people with softer frames.  It still is quite aggressive.  If it wobbles, shakes, or climbs to the moon, pull the master slider back until it doesn't do those kinds of things.

If you use a 250hz or 500hz RC link, check the relevant box, and the appropriate feedforward smoothing and jitter reduction settings will be applied.

Needs updating with Gary's rates, and some other minor modifications - not for merging right now.

### DIFF
--- a/presets/4.3/tune/whoop_justice.txt
+++ b/presets/4.3/tune/whoop_justice.txt
@@ -1,0 +1,246 @@
+#$ TITLE: Whoop race tune by Justice
+#$ FIRMWARE_VERSION: 4.3
+#$ CATEGORY: TUNE
+#$ STATUS: EXPERIMENTAL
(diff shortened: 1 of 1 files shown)


limonspb (commented):
VBat sag compensation, rc_smoothing, no deadband and "arm at any angle" technically have to be UNchecked, as they are not listed in defaults, and in the defaults.txt they are explicitly mentioned in the category " Other values may ONLY be included as un-checked options. Examples follow..."

KarateBrot (dismissed):

limonspb (approved):

PR #146, opened 2022-01-08
Update Tracer_CFV3.txt

xsnoopy:
Couldn´t find the  version of CFV3 in the Lua Script. Asked in the TBS facebook group and Raphael Pirker answered:

It will automatically negotiate this for you. If you use a recent version of betaflight and a version of tracer or crossfire 6.07 and up it will be v3

### DIFF
--- a/presets/4.3/rc_link/Tracer_CFV3.txt
+++ b/presets/4.3/rc_link/Tracer_CFV3.txt
@@ -7,7 +7,7 @@
 #$ DESCRIPTION: setup for TBS Tacer CFV3
 #$ DESCRIPTION: WARNING: make ABSOLUTELY SURE that the Hardware > ADC Filter is un-checked! and both radio and the TX unit are all up to date with the latest EdgeTX, OpenTX or FreedomTX
 #$ DESCRIPTION: Please use RC smoothing presets for flying style required. Note base settings are for racing, there is a check box at the end for freestyle.
-#$ DESCRIPTION: There are 2 current types of Crossfire protocol, CFV2 and CFV3 this can be checked in your TBS lua.
+#$ DESCRIPTION: There are 2 current types of Crossfire protocol, CFV2 and CFV3 (CFV3 with TBS Firmware V6.07 and newer for Crossfire and Tracer) this can be checked in your TBS lua.
 #$ DISCUSSION: https://github.com/betaflight/firmware-presets/pull/30
 
 feature RX_SERIAL


Zuldan:
@xsnoopy just so know, you only have CRSF v3.0 if your radio firmware has support for CRSF v3.0. At the moment OpenTx and EdgeTx do not support CRSF v3.0. The only radios that do support CRSF v3.0 are the Tango 2 and Mambo. Hopefully people on OpenTx and EdgeTx do not choose the v3.0 preset even if they have a module with v6.07.

KarateBrot (approved):

sugaarK:
@xsnoopy  mr @Zuldan is correct currently only the tango2 and mambo support the 11bit as there is no radio firmware available for other radios...  happy to address this when this changes but until then im closing this.. FYI I was the one whom did the testing with the V3 and BF...

xsnoopy:
@sugaarK i think just closing this is not addressing the issue. The description is not helpful and should be changed. A hint that edgeTX can't do V3 or like @Zuldan said shaw what radios can do V3 is in my opinion a minimum.

Zuldan:
@sugaarK how about changing 

`#$ TITLE: TBS Tracer CFV3`

to

`#$ TITLE: TBS Tracer CFV3 (TBS Mambo Only)`

and for the other preset

`#$ TITLE: TBS Crossfire CFV3 (TBS Tango 2 Only)`

then when TBS release the v3 PR to OpenTX/EdgeTX (and it’s merged) we can remove the limitation in the title. That way it’s pretty clear that the v3 presets are only for these radios at the moment. 

People are going to install these presets assuming that if they have the latest TBS module firmware they are running v3 end to end, which is incorrect.

sugaarK:
The real Way to fix the issue would be TBS actually making a PR for edge/OpenTX. That doesn’t see that  happening any time soon unfortunately. I was under the impression when I did this that TBS would move a LOT faster.

Zuldan:
Yep totally agree. It’s really annoying. I ask Trappy every livestream when he’s going to submit the PR and get the same response “OpenTx/EdgeTx changed the way everything communicates so now we have to rewrite the PR” with zero timeline. I guess that their Tango 2 and Mambo customers are happy and that’s all that really matters to them.

sugaarK:
@Zuldan totally tracer and crossfire both CFv3 ready as is BF it’s all tested and merged months ago..  could change the title..  I’d actually like to put them together into say 3 presets, TBS race, freestyle and long range

xsnoopy:
@Zuldan exactly what your described above is what happened to me. I thought I had v3 after asking in the TBS Facebook group and applied the v3 presets.

But turned out i didn't. Trappy made another comment on Facebook explaining it more. 

Let's fix the open source part of this (preset description) as we move faster and be more flexible ;)

xsnoopy:
@Zuldan changing the title is a good idea. I still would say we should change the wording of the description, because it's just not possible to see in the lua if you have v3 or v2. 

Only way to find that out is: right firmware of the TBS equipment, right radio firmware and BF 4.3 (that we use BF 4.3 when talking about presets should be imperative, since its my understanding that came with bf 4.3).

Just looking at the TBS lua is not doing the trick.

Zuldan:
@sugaarK that’s a really good idea. Would just 2 presets (Tracer/Crossfire) with 3 sub options each (Race/Freestyle/Cinematic) work? And maybe even with the presets, 150Hz/50hz/Dynamic, just have those as sub options to reduce the amount of Crossfire presets presented to the end user. I guess there are so many ways to do it. In my humble opinion, what ever is decided on, the I think the structure should be the same for all RC Links to keep it uniform.

@xsnoopy yeah TBS is making it really confusing for everyone. When EdgeTX release 2.7 (with full Tango 2/Mambo support) in March 2022, maybe we’ll have v3 then.

PR #153, opened 2022-01-09
replaced tbs rc link presets with race, freestyle and  longrange

sugaarK:
replaced the original tbs pre sets with race freestyle and longrange.. 
also created an ADC and radio firmware warning


sugaarK:
should be Better @Zuldan

limonspb (approved):

Zuldan:
@sugaarK it's beautiful!

image

PR #156, opened 2022-01-10
IRC Ghost preset

sugaarK:
presets in race, freestyle and longrange version.. 
and yes i have excluded mode that make no sense for the use


sugaarK:
there we go @limonspb

limonspb (approved):

PR #160, opened 2022-01-12
RC_LINK fixes: defaults.txt and typos

limonspb:
Current rc_link/defaults.txt is renamed to reset_to_defaults.txt
new defaults.txt has fewer parameters and can be used within rc_link presets.

Some typos.


sugaarK (changes requested):
hz should be Hz

limonspb:
> hz should be Hz

ok i think i fixed most of YOUR "hz" typos lol

sugaarK (approved):

presets/4.3/rc_link/defaults.txt
```diff
@@ -1,77 +1,13 @@
...
 #$ CATEGORY: RC_LINK
 #$ STATUS: OFFICIAL
 #$ KEYWORDS: defaults, reset, rc link, rc_link
 #$ AUTHOR: Betaflight
-#$ DESCRIPTION: Resets all RC Link related parameters to 4.3 defaults
-# Warning: Resets settings for ALL protocols
-# Note: Does NOT reset serial port settings or Rx Link mode (TAER / AETR etc)
+#$ DESCRIPTION: Resets RC_LINK feedforward smoothing/jitter/averaging parameters and RX provider protocol.
 
 set serialrx_provider = SPEK1024
```
ctzsnooze, line 9: If this file is about feedforward settings, we don't need this line to be included, since the kind of serial link to be used is irrelevant.

presets/4.3/rc_link/defaults.txt
```diff
@@ -1,77 +1,13 @@
-#$ TITLE: Default 4.3 RC link settings
+#$ TITLE: Reset RC link-feedforward related settings
 #$ FIRMWARE_VERSION: 4.3
 #$ CATEGORY: RC_LINK
 #$ STATUS: OFFICIAL
 #$ KEYWORDS: defaults, reset, rc link, rc_link
```
ctzsnooze, line 5: The keywords should include the word feedforward

presets/4.3/rc_link/TBS_race.txt
```diff
@@ -1,13 +1,13 @@
...
+#$ KEYWORDS: rc link, tracer, TBS, cfv2, 250Hz, crossfire, racing, 150Hz, CRSF
 #$ AUTHOR: sugarK
-#$ DESCRIPTION: setup fo the various TBS radio links via CRSF protocol for racing
-#$ DESCRIPTION: WARNING: make ABSOLUTELY SURE that the Hardware > ADC Filter is un-checked! and both radio and the TX unit are all up to date with the latest EdgeTX, OpenTX or FreedomTX
-#$ DESCRIPTION: There are 2 current types of Crossfire protocol, CFV3 if your using a TBS radio and CFV2 if your radio runs Open/EdgeTX
-#$ DESCRIPTION: Dynamic Crossfire or locked is not recommended for racing, locking at 150hz is preferred
+#$ DESCRIPTION: Setup fo the various TBS radio links via CRSF protocol for racing.
+#$ DESCRIPTION: WARNING: make ABSOLUTELY SURE that the Hardware > ADC Filter is un-checked! and both radio and the TX unit are all up to date with the latest EdgeTX, OpenTX or FreedomTX.
+#$ DESCRIPTION: There are 2 current types of Crossfire protocol, CFV3 if your using a TBS radio and CFV2 if your radio runs Open/EdgeTX.
+#$ DESCRIPTION: Dynamic Crossfire or locked is not recommended for racing, locking at 150Hz is preferred.
```
ctzsnooze, line 10: This line makes no sense.
It says 'or locked is not recommended' then says 'locking at 150hz is preferred.
Maybe they meant to say, "Dynamic or un-locked Crossfire modes are not recommended..."

presets/4.3/rc_link/IRC_race.txt
```diff
@@ -6,7 +6,7 @@
 #$ AUTHOR: sugarK
 #$ DESCRIPTION: Setup for the various ImmersionRC modes via GHST protocol for racing.
 #$ DESCRIPTION: WARNING: Make ABSOLUTELY SURE that the Hardware > ADC Filter is un-checked! and both radio and the TX unit are all up-to-date with the latest EdgeTX, OpenTX or FreedomTX.
-#$ DESCRIPTION: For racing it makes sense to use one of the 250/500Hz modes which the 'race' mode is only 166hz.
+#$ DESCRIPTION: For racing it makes sense to use one of the 250/500Hz modes which the 'race' mode is only 166Hz.
```
ctzsnooze, line 9: I don't understand this line.  The first part is true: "For racing it makes sense to use one of the 250/500Hz modes" is correct.  The second part , "which the 'race' mode is only 166Hz." seems to me too make no sense and should be deleted, or updated to say whatever was intended.

presets/4.3/rc_link/elrs_150hz.txt
```diff
@@ -1,10 +1,10 @@
-#$ TITLE: ELRS 150hz
+#$ TITLE: ExpressLRS 150Hz
 #$ FIRMWARE_VERSION: 4.3
 #$ CATEGORY: RC_LINK
 #$ STATUS: OFFICIAL
-#$ KEYWORDS: ELRS, rc, link, rc_link, 150hz
+#$ KEYWORDS: ELRS, rc, link, rc_link, 150Hz, express
```
ctzsnooze, line 5: good addition, thanks

presets/4.3/rc_link/IRC_race.txt
sugaarK: should be while 'Race' mode is only 166Hz

ctzsnooze: maybe `, because the 'race' mode is only 166Hz.`

sugaarK: That works

presets/4.3/rc_link/defaults.txt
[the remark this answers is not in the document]
limonspb: Pretty much every one of yours and Zak RC_LINK presets are setting this value, so I think if you OK with setting them in the RC_LINK presets, you should be ok also with it in the defaults file.

presets/4.3/rc_link/TBS_race.txt
limonspb: Yeah that's not my line, I just fixed typos :)
What exactly it should say?

limonspb:
> @limon I made some small suggestions. Thank you for looking at this so closely.

Thanks for looking! I commented back on some. With the rest of them I agree and will fix.

presets/4.3/rc_link/TBS_race.txt
sugaarK: `#$ DESCRIPTION: Dynamic Crossfire or locked 50Hz is not recommended for racing, locking at 150Hz is preferred. `

presets/4.3/rc_link/IRC_race.txt
ctzsnooze: `#$ DESCRIPTION: For racing it makes sense to use one of the 250/500Hz modes, because the 'race' mode is only 166Hz.`

presets/4.3/rc_link/defaults.txt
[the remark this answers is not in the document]
ctzsnooze: Yes, in an RC_LINK preset, there is a need to set the serial protocol correctly, so that the serial data stream is interpreted correctly.

However, there is no need to 'undo' this with a default, beforehand, because every RC_Link preset will always set this, or the link won't work.  It won't matter what the value is beforehand, it will need to be over-written.

Let's imagine a user who had FrSky and upgraded to ELRS.  The ELRS RC_Link Preset will force the serial provider to CRSF.  It doesn't matter if that Preset loaded a default which converted their prior FrSky serial provider to Spek1024, it will be changed to CRSF.  

The only time a default to Spek1024 would make any difference would be if we had a Spek1024 Preset.  That Preset, and only that Preset, would not need to change serial Rx provider.  

Spek1024 is an archaic Rx protocol, and frankly it makes no sense that it remains our default firmware protocol.  We would be best off setting the firmware default to a more common protocol like CRSF, especially as ELRS gains traction.  

However, it really is not necessary to set this in the reset file, and this line should be removed.

sugaarK (dismissed):

presets/4.3/rc_link/defaults.txt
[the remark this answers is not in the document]
limonspb: ok i can agree here :)

presets/4.3/rc_link/IRC_race.txt
limonspb: modified it like that:
#$ DESCRIPTION: 250Hz and 500Hz modes have less delay than the 'race' mode, which is only 166Hz.
Because 250Jz and 500Hz gives less latency regardless if thats for racing or not :)

ctzsnooze (approved):

PR #165, opened 2022-01-16
Added fpvian_basher_freestyle preset

FPVian:


FPVian:
Will resubmit without index changes

PR #168, opened 2022-01-17
Southwest Spec Racing Tunes-  SWPR GTB339 and TexasSpec (FreedomSpec)

TehllamaFPV:
This includes 4 preset files, with community and searchable tagging.

First preset is the SWPR GTB 339 Multi-Voltage tune.  'Only' tested on 27 quads now, works on TP3 layouts as well.  Auto-pulls 2S or 3S tune, and achieves remarkably solid performance from such a budget-friendly BNF/PNP setup.

Then is the family of FreedomSpec/TexasSpec setups.  All of these are intended to enable pilots to simply plug in a 3S 2200mAh pack and receive spec racing levels of performance if correct KV in the drop-down is selected, but provide a conservative open racing tune in the nominally correct voltage ranges for open spec.
The 1700-2250KV spec setup also includes the 'BMS Spec' configuration of flying 4S on a 1750KV quad, and performance in that configuration is pretty excellent.  These quads do not consistently fly well on 3S, therefore those tunes are omitted.
Current known limitations are that the 1900-2300KV preset lacks a 4S option (and can have minor issues with a 4S pack if the 3S profile is selected - usually fine, but did experience one flyaway over a sample size of 15 when testing), and that the higher kv 2350-2750KV option deliberately lacks a 6S option (i.e. expect failures, or at least 'danger to manifold' levels of performance, especially if the 3S profile is selected when plugging in a 6S pack.  I have tested it with a 5S pack and performance is solid up to 2650/2750KV, but without automatic behavior the 3S profile existing is a practical danger when user behavior is incorrect.
Warnings are present (and probably a bit heavy-handed) for these configurations.

Thus far I have 'only' tested these tunes on a combined 68 quads, from 5 builders (and two OEMs for BNF configurations), in varying states of frame degradation, but performance has been solid with cool motors even in 37°C temperatures with maximum voltage and WOT applied.


limonspb:
i will check in mode details later.
A quick look:
#$ STATUS has to be EXPERIMENTAL for all new presets unless they are very simple.
Later you can change that to COMMUNITY when more poeple will try that from the Configurator.

Slack links for the discussion are not very good. Slack removes old threads. We usually leave the link to the PR.

If you updating the filter values yu'd better include defaults for filters, not just TUNE. TUNE defaults don't have filter values.

including defaults for does not reset all profiles, only the current one. So i would include defaults after every profile swap.

because you are switching profiles, i would make this preset in OTHER category.

Because you are using sliders, you might want to use command "simplified_tuning apply". In CLI the sliders don't automatically recalculate PIDs/filters.

TehllamaFPV:
Definitely can fix the Filter default INCLUDE, since I don't explicitly alter every parameter in there.  Will fix that along with making the filtering includes optional.
Definitely need to fix the default per-profile defaults (I think I had been running a locally modified set of defaults in my branch to overwrite defaults in all profiles) - either way, especially since the Profile 3 for 3S packs is spicy, that's a nontrivial omission.
Will increment with additions to drop-down checkbox organization as well.

The Slack discussion links are old, and I just pointed them to the Presets channel - if that's OBE, I'll just remove them.

I can nudge them to experimental - I don't think any of the tune presets so far (save SugarK's Karate Race) have the same test sample size.  
I preferred the behavior where the sliders have to be activated (but do appear), but I can also add the slider apply.  Whichever is stylistically better as a template, I'm good with.

I'd still argue that these are TUNE presets - I can put the filters as included options (checked by default), the only parameters outside of the pure TUNE ones are the TPA adjustments, which can be set with the same drop-down box behaviors (or omitted if desired).

ctzsnooze:
These look good.  My comments:

- I don't know about `simplified_tuning apply` when multiple profiles are loaded.  Please make sure that each profile has numeric values for all slider controlled parameters.  For example, if Dmax is active, be sure to specify a value for D and for Dmax.  

- At the end of the list I assume you want to set the current profile and rates mode, and apply the sliders for that.

- For the filter settings, using option_group headers can be very useful to organise the options.  For instance, you can provide an option header 'Filter options' ....

- If these presets change TUNE parameters only (eg those reset by the TUNE default file), and all the other non-TUNE changes are 'options', then the preset can be called a TUNE.  In this case none of those options can be set by default.  The user needs encouraging to choose the most appropriate filter setup.

TehllamaFPV:
I need to add the filter defaults as an include, since these all rely on a combination of filters and PID values to fly appropriately.  Can do the same on grouping options, but I really want to insist that the filter profiles are requisite for making these tunes viable as intended (basically overly conservative but achieving comparable phase latency to the spicy karate tunes).

I can do a ton of regression testing to see what performance deltas may be present with default/noisy/karate type filtering on the same tunes, but I'm hesitant to do that because I prefer the validation heritage of the rest of the tune knowing that I can fly delaminated garbage at wide open throttle with a big sample size combining the PIDs/Tune, and so far this has been a drop-in fix even though it colors outside the lines completely (modifying TPA and filters), altering parameters that adjust flight performance... but not having the filtering defaults is a potential liability.  My personal argument is that if a pilot wants different filtering behavior, they can run a different filtering profile manually after the fact, but I'd be happy to can put that as a default checked option in its own option group.

The intended users for this preset would be people who already have a race quad of some description, and just want to be able to experience spec racing performance by installing an $8 battery, and there are some filtering adjustments one can only get away with on 3S reliably, but those are also contingent on a particular gyro filtering scheme as well that isn't as profile dependent, hence wanting them as a complete package... and this phenomenon is worse on motor output limited high-KV setups with 3S packs, although I don't know how much because I've only ran ~100 packs through those with this final setup.

I had problems with simplified_tuning_apply, might have been just with using multiple profiles where I had legacy typos - without that, the configurator and OSD all show the correct slider values, but it more reliably applied the tune in all cases - for end users to go and modify, enabling sliders retains these PID values, and I think the missing ones are where it's actually the stock D gains in places.  Will regression test those behaviors, see if it'll take on all three profile, since these are intended to be base/starter tunes in reality.

Pending update - I need to rebase my local branches to get node.js to work right - sorry about those delays.

TehllamaFPV:
@Limon I think this PR is ready to merge.

Added the filter default include, marked all as Experimental, and conducted additional testing on these.  Really my only glaring complaint is not having a way to reliably force a particular profile on the 3, 5, 6 cell tune setups if a 4S pack is plugged in.

I've tried on the simplified_tuning_apply, but with multiple profiles and/or using any values beyond the non-expert mode sliders, it still just quits, so I don't know how to force the configurator to enable sliders as the default behavior... they are there, with the right values. Same with any of the non-specified values (I think in three cases the sliders deposit me on a default D gain value, which this reliably inherits from the #% INCLUDE of the default tune.

I started making a version that pushes everything not part of the TUNE defaults into options, but honestly I don't want to test all of these tunes on default filtering, and wouldn't want anybody else to do likewise... so leaving it in the 'Other' bin works for now.

presets/4.3/other/Tehllama_5in_2350-2750KV-Freedom_3S-4S-5S_Race_Tune.txt
```diff
@@ -0,0 +1,245 @@
...
+set motor_output_limit = 78
+set simplified_master_multiplier = 95
+#$ OPTION END
+
+#$ OPTION BEGIN (UNCHECKED):  2650KV Motor Output Limiting (3S and 5S)
+profile 1
+set motor_output_limit = 94
+profile 2
+set motor_output_limit = 74
+set simplified_master_multiplier = 90
```
limonspb, line 236: this won't take any effect without simplified_tuning apply following after.

presets/4.3/other/Tehllama_5in_2350-2750KV-Freedom_3S-4S-5S_Race_Tune.txt
```diff
@@ -0,0 +1,245 @@
...
+#$ FIRMWARE_VERSION: 4.3
+#$ CATEGORY: OTHER
+#$ STATUS: EXPERIMENTAL
+#$ KEYWORDS: 533, HighKV, FreedomSpec, Spec, Race, Llama, MultiVoltage
+#$ AUTHOR: Daniel Appel / Tehllama
+#$ DESCRIPTION: Built for EMULATING 533 FreedomSpec and similar builds on bidir DShot ESCs at 48kHz PWM with 23°/MedHigh timing
+#$ DESCRIPTION: 3S (2200mAh) will run a spec tune.  4S and 5S will run a conservative open race tune.
+#$ DISCLAIMER: This tune will auto-select profiles based on battery voltage at plugin, but only for 3S, 4S and 5S packs. Extensive testing has been done across 54+ builds, however not every craft will run best on this tune.
+#$ WARNING: Use at your own risk.  NO 6-cell tune or throttle limit is provided with this tune.  Plugging in a 6S pack with 3S profile active will likely cause damage.  
+#$ DISCUSSION: https://app.slack.com/client/T221J1F6J/C0228HK2MD4/
```
limonspb, line 11: discussion link on slack is not a nice idea because this discussion will go away after some time on free accounts.

index_hash.txt
```diff
@@ -1 +1 @@
-62e24e16090268ccc53e701dfd411f21036291c2a5320d6ea056b761e2813e55
```
limonspb, line 1: plz undo changes for this file and for index.json

presets/4.3/other/Tehllama_5in_1700-2250KV_4S-5S-6S_Race_Tune.txt
```diff
@@ -0,0 +1,234 @@
...
+#$ OPTION END
+
+#$ OPTION BEGIN (UNCHECKED):  2050KV 6S Motor Output Limit
+profile 0
+set motor_output_limit = 98
+profile 2
+set simplified_master_multiplier = 120
+#$ OPTION END
+
+# save configuration
```
limonspb, line 234: no need for this line

presets/4.3/other/Tehllama_5in_2350-2750KV-Freedom_3S-4S-5S_Race_Tune.txt
```diff
@@ -0,0 +1,245 @@
...
+set tpa_breakpoint = 1250
+
+rateprofile 5
+set tpa_rate = 72
+set tpa_breakpoint = 1250
+
+rateprofile 0
+#$ OPTION END
+
+#$ OPTION BEGIN (CHECKED): Apply TPA Settings to current rateprofile only
```
limonspb, line 200: why you want TPA for the current profile only?

limonspb (changes requested):

presets/4.3/other/Tehllama_5in_2350-2750KV-Freedom_3S-4S-5S_Race_Tune.txt
TehllamaFPV: Realistically a somewhat vestigial option - but I wanted to provide a choice where if somebody had 'race' rates on a setup, and different cinematic rates, it would leave every other profile alone.  I can't actually imagine a use case where that is relevant, but wanted to leave an option.

index_hash.txt
TehllamaFPV: Unsure how to actually do - will try.

presets/4.3/other/Tehllama_5in_2350-2750KV-Freedom_3S-4S-5S_Race_Tune.txt
TehllamaFPV: That's the part I've been struggling with - I can't even get a reliable CLI command to actually perform the simplified_tuning apply to run.

presets/4.3/other/Tehllama_5in_2350-2750KV-Freedom_3S-4S-5S_Race_Tune.txt
TehllamaFPV: I finally remembered why - whenever it applies rateprofile changes to ALL rate profiles, it drops the user back on RateProfile 0, because of how it works.  If that is an undesired behavior, users can select 'current profile only'

presets/4.3/other/Tehllama_5in_2350-2750KV-Freedom_3S-4S-5S_Race_Tune.txt
limonspb: @TehllamaFPV 
simplified_tuning apply

and in this case you don't even need to set PID values, only sliders.

TehllamaFPV:
Update: I now have the discussion link corrected (points to this PR discussion), moved all of the non-Tune options to options (checked by default), and implemented a fallback filter array options and cleaned up the dynamic idle options.

My other question would be what the benchmark would need to be for these to shift from Experimental to Community or Official - I'm not honestly troubled either way, I just want these in the master branch so people with craft that do not respond well to the Karate tune have this as an immediate fallback option to test if their birds just need more 'accommodating' filters.

presets/4.3/tune/Tehllama_GTB339_3in_SWPR-Spec_2S-3S_Race_TUNE_Options.txt
```diff
@@ -0,0 +1,173 @@
...
+#$ FIRMWARE_VERSION: 4.3
+#$ CATEGORY: TUNE
+#$ STATUS: EXPERIMENTAL
+#$ KEYWORDS: GTB339, Toothpick, Spec, Race, Llama, SWPR, MultiVoltage
+#$ AUTHOR: Daniel Appel / Tehllama
+#$ DESCRIPTION: This tune is intended for GTB339 3in 1105-5500KV Racers on on bidir DShot ESCs at 48kHz PWM with MedHigh timing. 
+#$ DESCRIPTION: These fly best on 2S-3S batteries in the 450-650mAh capacity range, resulting in an AUW to 95-105g on these motors.  Works with biblade and triblade props (e.g. Gemfan 3016/3018, HQ 3x1.8x3/3x1.5x3, etc.)
+#$ DISCLAIMER: This tune will auto-select profiles based on battery voltage at plugin for 2S and 3S packs.
+#$ WARNING: Requires 12-Pole Motors!!! Use at your own risk.  Extensive testing has been done across 37+ builds, however not every craft will run best on this tune.  13A ESCs may experience failures on 3S batteries.
+#$ DISCUSSION: https://app.slack.com/client/T221J1F6J/C0228HK2MD4/
```
limonspb, line 11: wrong link

presets/4.3/tune/Tehllama_5in_2350-2750KV-Freedom_3S-4S-5S_Race_TUNE_Options.txt
```diff
@@ -0,0 +1,286 @@
...
+
+set simplified_master_multiplier = 65
+set simplified_i_gain = 85
+set simplified_d_gain = 110
+set simplified_pi_gain = 125
+set simplified_dmax_gain = 85
+set simplified_feedforward_gain = 150
+set simplified_pitch_d_gain = 110
+set simplified_pitch_pi_gain = 110
+set simplified_dterm_filter = OFF
```
limonspb, line 93: some filter changes in these presets are not covered with #$ include filter defaults. For example these ones.

You are modifying filter settings here, but you do not apply any filter defaults before that. So technically at this point you might end up with any random filters + your few filter changes.

Please check the same for the rest of the presets.

presets/4.3/tune/Tehllama_GTB339_3in_SWPR-Spec_2S-3S_Race_TUNE_Options.txt
```diff
@@ -0,0 +1,173 @@
...
+# profile 1: 2S
+profile 1
+
+set vbat_sag_compensation = 77
+set anti_gravity_gain = 4400
+set iterm_rotation = ON
+set iterm_relax = RPY
+set iterm_relax_cutoff = 37
+set yaw_lowpass_hz = 125
+set p_pitch = 82
```
limonspb, line 65: all these direct PID values are going to be overwritten by simplified_tuning apply, so you don't have to mention them.

presets/4.3/tune/Tehllama_GTB339_3in_SWPR-Spec_2S-3S_Race_TUNE_Options.txt
```diff
@@ -0,0 +1,173 @@
+#$ TITLE: 3in Lightweight Race GTB339 3S/2S Tune
+#$ FIRMWARE_VERSION: 4.3
+#$ CATEGORY: TUNE
```
limonspb, line 3: i still don't agree this is a TUNE preset. I think this has to be in OTHER.
With the TUNE presets people are guaranteed not to have any left overs if they try different TUNE presets.

Your presets have the following left overs:
auto_profile_cell_count
changing profile
motor_poles
vbat_sag_compensation 

all this is not alloowed in the TUNE unless added in the UNCHECKED options.

OTHER category gives way more flexibility and people will understand they will have leftovers.
Also, it is good to mention these leftovers in the description.

I also think this PR is getting too complicated. Let's work on one preset at the time if they are that complex.
THanks!

presets/4.3/tune/Tehllama_GTB339_3in_SWPR-Spec_2S-3S_Race_TUNE_Options.txt
```diff
@@ -0,0 +1,173 @@
...
+
+rateprofile 0
+#$ OPTION END
+
+#$ OPTION BEGIN (UNCHECKED): Apply TPA Settings to current rateprofile only
+set tpa_rate = 72
+set tpa_breakpoint = 1350
+#$ OPTION END
+
+# save configuration
```
limonspb, line 173: this line is still here

limonspb (changes requested):

presets/4.3/tune/Tehllama_GTB339_3in_SWPR-Spec_2S-3S_Race_TUNE_Options.txt
TehllamaFPV: The auto-profile-select is somewhat integral to these, but I can definitely add the VbatSagComp into the options bin for each.
The motor pole forcing is only for specific spec racing builds (12P 1105 5500KV Mamba or 1404 TinyTurners)... I can wrap them in an option too.

TehllamaFPV:
Made a large update (not really any new content, just reorganized everything into selection bins.

Thanks for your feedback, I hadn't really tested any configuration edge cases on this, and definitely had some copy-paste errors in there.

I definitely would insist on most of the defaults to checked (motor poles, filtering, etc.), primarily because I'd rather have depth of testing with the following setups, I think it would take me a few permutation to try each of these with partial box deselections, realistically I have probably a thousand flights on the current default checkboxes, and only a few without.

TehllamaFPV:
Rebuilding as 4 PRs for each tune.

PR #171, opened 2022-01-18
Switch to GitHub Pages as deployment / hosting method

PhilippMolitor:
# Problem

The current setup requires for every change to push a compiled index file and the corresponding preset assets to some S3 instance. When other contributors want to provide their own URLs for custom repositories it means that they too have to provide some form of S3 for hosting, which is additional effort, and it's also pretty complicated for someone who just wants to publish their own preset library.

If I misunderstood something about that, please let me know.

# Proposed solution: use GitHub Pages to host the index file

I created a working prototype on my fork of this repository.
**_Note that to replicate this, you have to enable GitHub Pages in the repository settings. This is currently hardcoded to the branch "index"._**
You can see the hosted version here: [https://philippmolitor.github.io/firmware-presets/](https://philippmolitor.github.io/firmware-presets/)

Changes that were made:

- [x] refactor and cleanup of the javascript indexer
- [x] improved on absolute and relative path handling
- [x] refactored github actions:
  - [x] on push to master/main branch, check create index, copy presets/misc folder, deploy to github pages
  - [x] on push to any other branch and pull requests, only checks will be executed. This is useful for checks on pull requests.
- [ ] Update README.md with new instructions on how to set this up
- [x] Check if the betaflight configurator can handle this without any issues
- [ ] When implementing this, the betaflight configurator must change the source URL to the github pages URL.

# Resulting improvements

- Refactored/simplified the JavaScript code
- Formatted the Javascript code properly (with the help of [https://prettier.io/](https://prettier.io/))
- The `index.json` and `index_hash.txt` now don't live on the master branch, but only on the Pages branch. This should reduce complexity and also should result in fewer commits that don't relate to actual changes of the presets significantly
- Contributers will have an easier time getting started
- Contributers can now easily host their own 3rd party preset repositories for use with the configurator


limonspb:
@PhilippMolitor thanks for the PR.

A couple of moments:

> When other contributors want to provide their own URLs for custom repositories it means that they too have to provide some form of S3 for hosting, which is additional effort, and it's also pretty complicated for someone who just wants to publish their own preset library.

I think that is not true. When the other contributors want to provide their own URL they don't need S3. The configurator can eat straight GitHub links + branch names. You can try for example:
**link:** https://github.com/limonspb/firmware-presets
**branch:** option_group
image

The only requirement for such branches is that the author has to run indexer.js on them and add/commit a new index.json

Automatic run of indexer on push to master/main - that can be useful.

About indexer refactoring - unfortunately, i think your editor replaced all tabs with two spaces instead of 4.
So it looks like EVERY js file has changed and it is impossible to review of what has actually changed.

PhilippMolitor:
I tried moving files and setting formatter options so it has comparable elements again - tho it is not perfect, as a lot of formatting (especially breaking up lines into multiple lines because of their length) happened.
If you find those changes to drastic please feel free to close this PR.
I see that it is not that useful regarding that i totally overlooked the option to use git repos instead of the S3 source.

If you still want to compare changes, you'll see that figuring out done quite fast. Note that `index.js` and both workflow files have been completely rewritten, comparison does not make much sense here. `index.js` has a few explaining comments though, and the CI scripts should be short enough to be easy to understand - if you have questions, I am of course happy to help.

indexer/IndexContent.js
```diff
@@ -1,51 +1,61 @@
...
-                }
-            }
-        }
-
-        result = [...result];
-        result.sort((a, b) => a.localeCompare(b, undefined, {sensitivity: 'base'}));
-        return result;
-    }
-
+"use strict";
```
PhilippMolitor, line 1: this is all auto formatting. no real changes here.

limonspb:
not needed as per PR author

PR #174, opened 2022-01-29
AyyyKayyy 1S whoop race tune

limonspb:
1S whoop brushless tune for acro racing.

AyyyKayyy - Whooptopia winner 2019, 2020, and 2021. Whooptopia - biggest and the most intense whoop race in the USA.

This tune does not require RPM and works with regular BLHELI_S

### DIFF
--- a/presets/4.3/tune/whoop_ayyykayyy.txt
+++ b/presets/4.3/tune/whoop_ayyykayyy.txt
@@ -1,0 +1,130 @@
+#$ TITLE: Whoop race tune by AyyyKayyy
+#$ FIRMWARE_VERSION: 4.3
+#$ CATEGORY: TUNE
+#$ STATUS: EXPERIMENTAL
+#$ KEYWORDS: whoop, 65mm, 1S, race, acro, tune, AyyyKayyy, meteor, whooptopia, brushless
+#$ AUTHOR: AyyyKayyy - Anthony Knight
+#$ DISCUSSION: https://github.com/betaflight/firmware-presets/pull/174/
+
+#$ DESCRIPTION: 65mm 1S brushless whoop acro tune for racing.
+#$ DESCRIPTION:
+#$ DESCRIPTION: Recommended gear this tune works the best with:
+#$ DESCRIPTION: 22000 - 25000kv motors, Meteor65 frame or simimlar, HQ tri-blades props.
+#$ DESCRIPTION: Works great with BBBLV3, BetaAIO 1s or HappyModel Diamond FC.
+#$ DESCRIPTION:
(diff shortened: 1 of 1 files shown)


presets/4.3/tune/whoop_ayyykayyy.txt
```diff
@@ -0,0 +1,133 @@
...
+set iterm_relax_cutoff = 20
+
+# -- iTerm rotation
+set iterm_rotation = ON
+
+# -- TPA (default)
+set tpa_breakpoint = 1250
+
+# -- Feedforward
+set feedforward_transition = 21
```
sugaarK, line 64: We don’t recommenced using ff trans in 4.3 it’s been superseded by jitter_reduction

sugaarK (commented):
Not sure about using ff transition jitter reduction is the preferred  way to do this now

presets/4.3/tune/whoop_ayyykayyy.txt
limonspb: > Not sure about using ff transition jitter reduction is the preferred way to do this now

Yeah agree. I will ask. But if the author says that's intentional, we should let it be I think.

limonspb: @sugaarK  good catch.
Removed

limonspb:
> Not sure about using ff transition jitter reduction is the preferred way to do this now

good catch.
Removed

presets/4.3/tune/whoop_ayyykayyy.txt
```diff
@@ -0,0 +1,130 @@
...
+set iterm_relax_cutoff = 20
+
+# -- iTerm rotation
+set iterm_rotation = ON
+
+# -- TPA (default)
+set tpa_breakpoint = 1250
+
+# -- Antigravity
+set anti_gravity_gain = 0
```
sugaarK, line 64: You sure you want  antigravity set to zero in 4.3, It works different to 4.2

sugaarK (changes requested):

presets/4.3/tune/whoop_ayyykayyy.txt
limonspb: > You sure you want antigravity set to zero in 4.3, It works different to 4.2

let me ask lol

limonspb: > You sure you want antigravity set to zero in 4.3, It works different to 4.2

@sugaarK  i just flown this tune, its a real fire. it does not need AG :)

sugaarK (approved):

PR #181, opened 2022-02-06
update ctzsnooze's presets for option groups

ctzsnooze:
Uses option groups to more effectively group the options.

Minor updates.

Option groups are awesome!


limonspb:
@ctzsnooze could you please remove index and index_hash from the PR. Other than that looks good!

limonspb (approved):

PR #189, opened 2022-02-27
Remove clean filter option from justice's whoop preset

ctzsnooze:
This PR removes the 'Clean' filter option from Justice's whoop preset, since many people were choosing it on builds that couldn't handle it, and not knowing why things didn't work out so great.

A clean filter Preset can still be applied as a second independent action, if desired.

Option groups have also been used to organise the options a bit better.

### DIFF
--- a/presets/4.3/tune/whoop_justice.txt
+++ b/presets/4.3/tune/whoop_justice.txt
@@ -106,10 +106,6 @@ set dshot_idle_value = 600
 #$ OPTION_GROUP BEGIN: Filters
 
 # Filter options
-#$ OPTION BEGIN (UNCHECKED): FILTER: RPM enabled, clean build
-#$ INCLUDE: presets/4.3/filters/default_rpm_clean.txt
-set dyn_idle_min_rpm = 50
-#$ OPTION END
 
 #$ OPTION BEGIN (UNCHECKED): FILTER: RPM enabled, normal build
 #$ INCLUDE: presets/4.3/filters/default_rpm_normal.txt
@@ -129,7 +125,7 @@ set dyn_idle_min_rpm = 50
 
 # -- Fast Rx Link Options --
 
-#$ OPTION_GROUP BEGIN: RC Links
+#$ OPTION_GROUP BEGIN: Your RC link speed, if above 150
 
 
 #$ OPTION BEGIN (UNCHECKED): RC_LINK 250Hz
@@ -146,7 +142,12 @@ set feedforward_jitter_factor = 4
 
 #$ OPTION_GROUP END
 
-#$ OPTION_GROUP BEGIN: Tune
+
+#$ OPTION_GROUP BEGIN: Tune options, check only one
+
+#$ OPTION BEGIN (UNCHECKED): Normal tune (normal builds)
+# do nothing, it's already applied
+#$ OPTION END
 
 #$ OPTION BEGIN (UNCHECKED): Spicy tune (clean builds only)
 # -- PID values --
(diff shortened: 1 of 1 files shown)


presets/4.3/tune/whoop_justice.txt
```diff
@@ -218,48 +215,61 @@ set simplified_master_multiplier = 100
 
 #$ OPTION_GROUP END
 
+
+#$ OPTION_GROUP BEGIN: Recommended settings, please check all of them
+
 # -- RC_Smoothing --
-#$ OPTION BEGIN (UNCHECKED): Default RC_smoothing (recommended)
+#$ OPTION BEGIN (UNCHECKED): Default RC_smoothing
```
KarateBrot, line 222:
```suggestion
#$ OPTION BEGIN (CHECKED): Default RC_smoothing
```

presets/4.3/tune/whoop_justice.txt
```diff
@@ -218,48 +215,61 @@ set simplified_master_multiplier = 100
...
 #$ INCLUDE: presets/4.3/rc_smoothing/race.txt
 #$ OPTION END
 
 # -- VBat warning threshold --
-#$ OPTION BEGIN (UNCHECKED): Default warning voltage (recommended)
+#$ OPTION BEGIN (UNCHECKED): Default warning voltage
```
KarateBrot, line 227:
```suggestion
#$ OPTION BEGIN (CHECKED): Default warning voltage
```

presets/4.3/tune/whoop_justice.txt
```diff
@@ -218,48 +215,61 @@ set simplified_master_multiplier = 100
...
 set vbat_warning_cell_voltage = 350
 #$ OPTION END
 
 # -- No deadband --
-#$ OPTION BEGIN (UNCHECKED): No deadband (recommended)
+#$ OPTION BEGIN (UNCHECKED): No deadband
```
KarateBrot, line 232:
```suggestion
#$ OPTION BEGIN (CHECKED): No deadband
```

presets/4.3/tune/whoop_justice.txt
```diff
@@ -218,48 +215,61 @@ set simplified_master_multiplier = 100
...
 set deadband = 0
 set yaw_deadband = 0
 #$ OPTION END
 
 # -- Startup and arming --
-#$ OPTION BEGIN (UNCHECKED): Arm at any angle (recommended)
+#$ OPTION BEGIN (UNCHECKED): Arm at any angle
```
KarateBrot, line 238:
```suggestion
#$ OPTION BEGIN (CHECKED): Arm at any angle
```

presets/4.3/tune/whoop_justice.txt
```diff
@@ -218,48 +215,61 @@ set simplified_master_multiplier = 100
 
 #$ OPTION_GROUP END
 
+
+#$ OPTION_GROUP BEGIN: Recommended settings, please check all of them
```
KarateBrot, line 219:
```suggestion
#$ OPTION_GROUP BEGIN: Recommended settings
```

KarateBrot (commented):
We could check the recommended settings by default, if this should be checked by the user anyways (I guess).

ctzsnooze:
@KarateBrot - it would be nice to pre-check all the recommended settings, but we have a 'rule' that we only pre-check values that are cleared by the defaults file.  So that they must be actively chosen, meaning that the person knows what extra changes were made and have some memory of doing them.  That's because those changes will not be reset by applying the tune 'defaults' file.

Justicefpv (commented):
Although it's not something I personally use I think setting the battery warning to the following is beneficial for 1s whoops. Anyone running older PH2.0 connectors will see their voltage warning pop up almost immediately with a warning when set at 350. Raising the max voltage to 440 stops slightly overcharged HV packs from reporting as 2s. 

` set vbat_max_cell_voltage = 440

  set vbat_min_cell_voltage = 320`

limonspb (approved):

PR #196, opened 2022-03-09
Update IRC Ghost presets

sugaarK:
this pr brings all the IRC Ghost presets up to the same page including rc smoothing settings for each type and style


sugaarK:
after discussion with @limonspb a better way has been found

PR #200, opened 2022-03-10
DJI-SBUS-FAST (Update to Add RC Smoothing Factors)

spatzengr:
Added RC Smoothing per request.  Snagged from ELRS 250hz by CTZ.


limonspb (changes requested):
Could you also change the description, as now it changes not just feedforward settings?
(Same as the other PR)
Also same as the other preset, the option names can be just "Racing, HD Freestyle, Cinematic" without mentioning feedforward.

limonspb (approved):

PR #201, opened 2022-03-11
UAV Tech 5" Preset

spatzengr:
UAV Tech 5" Preset

PID Tab:
image

Filters Tab:
image

48K PWM Option:
image

PR #202, opened 2022-03-11
UAV Tech 6/7in Preset

spatzengr:
UAV Tech 6/7in Preset

PID Tab:
image

Filters Tab:
image

48K PWM Option:
image


limonspb (changes requested):
Thanks for the preset!

Similar comments as in this PR, please have a look
https://github.com/betaflight/firmware-presets/pull/201

limonspb (approved):

PR #203, opened 2022-03-11
New RC_LINK presets for Ghost and TBS

limonspb:
Experimental presets based on generic values until we get better values from BlackBox recordings.

if this PR is not merged yet then:
add custom source in the configurator:
https://github.com/limonspb/firmware-presets
branch name: ghost_test

If PR is merged already then proceed with the following (using official preset source):

For testers, who are willing to provide BlackBox to improve RC_LINK presets for Ghost/Tracer/Crossfire:
1. Select RC mode you want to test in Ghost/Crossfire or Tracer
2. Select the existing RC_LINK preset for the mode you want to test.
3. Select Race or Freestyle option.
4. Set logging in the following way:
image
5. **Take props OFF** or don't even power the motors (make sure RX is powered)
6. "Arm" your drone.
7. start slow, do like 0.5hz roll moves. Stay inside one-half of the range and avoid hitting center or limits. Then slowly get faster and faster.
8. Then some ‘imaginary’ flips or rolls.  and some ‘pretend flying

Download blackbox and share it here.
Mention what mode you were testing and which preset did you apply.


limonspb:
logos:
Ghost-Logo

communityIcon_tkc6w2d3zxz01

ctzsnooze (approved):

Justicefpv:
These are with the latest 12bit code from Stepan

[GHOST_RACE500_12BIT_DB_FEEDFORWARD.txt](https://github.com/betaflight/firmware-presets/files/8239175/GHOST_RACE500_12BIT_DB_FEEDFORWARD.txt)
[GHOST_SOLID250_12BIT_DB_FEEDFORWARD.txt](https://github.com/betaflight/firmware-presets/files/8239176/GHOST_SOLID250_12BIT_DB_FEEDFORWARD.txt)

limonspb:
> These are with the latest 12bit code from Stepan
> 
> [GHOST_RACE500_12BIT_DB_FEEDFORWARD.txt](https://github.com/betaflight/firmware-presets/files/8239175/GHOST_RACE500_12BIT_DB_FEEDFORWARD.txt) [GHOST_SOLID250_12BIT_DB_FEEDFORWARD.txt](https://github.com/betaflight/firmware-presets/files/8239176/GHOST_SOLID250_12BIT_DB_FEEDFORWARD.txt)

Thanks!
which options did you use?

Justicefpv:
I used the race for both

joha95:
The include of the options (race/freestyle/...) overwrite the settings of the Ghost setting written before.

limonspb:
> The include of the options (race/freestyle/...) overwrite the settings of the Ghost setting written before.

could you please be more specific:
1. is that a problem in your opinion?
2. which settings exactly?

joha95:
If you select the ghost presets, for example solid250, the preset sets the ghost specific ghost settings, like FF smooth factor to 45, jitter to 6, etc. If you select the race option, it's includes the generic race250 which overwrite the smooth factor to 35, jitter to 4. On top it change the RC smooth factor to 25, which is the only setting that should be changed with the race option.

1. Yes I think it's a problem, because atm the link specific settings get overwritten (in this case ghost solid250)
2. Settings that get overwritten:
feedforward_averaging, feedforward_smooth_factor, feedforward_jitter_factor 

So my tip would be to set the link specific settings after the includes of the generic settings.

limonspb:
> If you select the ghost presets, for example solid250, the preset sets the ghost specific ghost settings, like FF smooth factor to 45, jitter to 6, etc. If you select the race option, it's includes the generic race250 which overwrite the smooth factor to 35, jitter to 4. On top it change the RC smooth factor to 25, which is the only setting that should be changed with the race option.
> 
> 1. Yes I think it's a problem, because atm the link specific settings get overwritten (in this case ghost solid250)
> 2. Settings that get overwritten:
>    feedforward_averaging, feedforward_smooth_factor, feedforward_jitter_factor
> 
> So my tip would be to set the link specific settings after the includes of the generic settings.

you are right that "feedforward_averaging, feedforward_smooth_factor, feedforward_jitter_factor" get overwritten by options.
That is expected. Depending on the purpose (race/freestyle/cinematic) these parameters will be different.

Solid250 requires different `feedforward_averaging, feedforward_smooth_factor, feedforward_jitter_factor`, depending on the flight style, not only rc_smoothing.

Also check this comment:
image

achmiral:
Hi @limonspb, is this TBS preset applicable for futaba radio?

limonspb:
> Hi @limonspb, is this TBS preset applicable for futaba radio?

For any radio, unless your radio does not support the desired update rate

syrekszymon:
Maybe You could add this to Crossfire 50Hz
```
set osd_link_quality_alarm = 70
set osd_RSSI_dbm_alarm = -100

```
and this, for 150Hz
```
set osd_link_quality_alarm = 40
set osd_RSSI_dbm_alarm = -100

```

PR #215, opened 2022-03-12
RPM off by Default

spatzengr:
The default for Betaflight (after the Reset button) is that the RPM filter is OFF.  However, the default Filter preset has the RPM filter enabled.  This change turns it OFF to be more consistent with the true default.

image

### DIFF
--- a/presets/4.3/filters/defaults.txt
+++ b/presets/4.3/filters/defaults.txt
@@ -34,6 +34,7 @@ set dyn_notch_max_hz = 600
 set gyro_filter_debug_axis = ROLL
 
 # -- RPM filtering --
+set dshot_bidir = OFF
 set rpm_filter_harmonics = 3
 set rpm_filter_q = 500
 set rpm_filter_min_hz = 100


limonspb (changes requested):
i see that by default after flashing it is like that:
image

but we certainly missing this one:
image

KarateBrot:
@limonspb Do we really want to switch off bidir dshot? There can be other features using RPM data so that would disable all features dependent on RPM data.

spatzengr:
> > @limonspb
> > but we certainly missing this one:
> > image
> 
> Do we really want to switch off bidir dshot? There can be other features using RPM data so that would disable all features dependent on RPM data, not just the RPM filter.

The fact is by default (after reset), it is off.  So our default.txt should reflect that.

limonspb:
> > @limonspb
> > but we certainly missing this one:
> > image
> 
> Do we really want to switch off bidir dshot? There can be other features using RPM data so that would disable all features dependent on RPM data, not just the RPM filter.

tough question. But im thinking about presets scenario:
1. You apply Karate TUNE with bidir=ON and RPM filters.
2.  For whatever reason you don't like it and decided to apply tune preset "B" where bdir=OFF
3. If we don't turn BDIR=OFF in the defaults, you will end up with preset "B" + RPM filters :)

limonspb:
So we either mandate app filters presets to specify BDIR ON/OFF or we add it OFF to defaults.
Im fine with both solutions...
Maybe first is even better?

KarateBrot:
If bidir is off by default this is fine.

bidir dshot is not connected to the RPM filter!!! It is actually disconnected but was for some reason intertwined with the RPM filter code.

If we really only want to disable the RPM filter, we MUST only `set rpm_filter_harmonics = 0` and `leave bidir unchanged` in case we need RPM data for other features. In the future (with more bidir dependent features) this will be even more relevant.

limonspb:
> If bidir is off by default this is fine.
> 
> bidir dshot is not connected to the RPM filter!!! It is actually disconnected but was for some reason intertwined with the RPM filter code.
> 
> In a hierarchy you can think of bidir DShot being the "master feature" and all features dependent on RPM data being "slave features". You can turn them on/off individually, but turning bidir DShot off in the motors tab turns them all off.
> 
> ```
>   Bidir DShot (on/off)
>     |
>     | - - - - RPM filter (on/off)
>     |
>     | - - - - Dynamic Idle (on/off)
>     |
>     | - - - - Feature 3 (on/off)
>     |
>       ...
> ```
> 
> If we really only want to disable the RPM filter, we MUST only `set rpm_filter_harmonics = 0` and **leave bidir unchanged** in case we need RPM data for other features. In the future (with more bidir dependent features) this will be even more relevant.
> 
> Right now only RPM filter and dynamic idle use bidir afaik, but I know that at least a third feature dependent on bidir might be added this year to BF 4.4.

But if we leave bidir unchanged, it means that after applying presets the users will have to set bdir on/off manually.

limonspb:
> If bidir is off by default this is fine.
> 
> bidir dshot is not connected to the RPM filter!!! It is actually disconnected but was for some reason intertwined with the RPM filter code.
> 
> In a hierarchy you can think of bidir DShot being the "master feature" and all features dependent on RPM data being "slave features". You can turn them on/off individually, but turning bidir DShot off in the motors tab turns them all off.
> 
> ```
>   Bidir DShot (on/off)
>     |
>     | - - - - RPM filter (on/off)
>     |
>     | - - - - Dynamic Idle (on/off)
>     |
>     | - - - - Feature 3 (on/off)
>     |
>       ...
> ```
> 
> If we really only want to disable the RPM filter, we MUST only `set rpm_filter_harmonics = 0` and **leave bidir unchanged** in case we need RPM data for other features. In the future (with more bidir dependent features) this will be even more relevant.
> 
> Right now only RPM filter and dynamic idle use bidir afaik, but I know that at least a third feature dependent on bidir might be added this year to BF 4.4.

Or even more weird scenario.
Imagine someone running bdir=ON but he wants to try default filtering again.
He opens presets, ally defaults. It will be defaults except BDIR will be ON for him with all the RPM filters. Cuz by default RPM harmonics is 3.

KarateBrot:
> Cuz by default RPM harmonics is 3.

Ok, I think there's the error. This worked back when bidir was intertwined with the RPM filter. But RPM harmonics should be 0 by default for BF 4.3 and onwards.

limonspb:
> > Cuz by default RPM harmonics is 3.
> 
> Ok, I think there's the error. This worked back when bidir was intertwined with the RPM filter. But RPM harmonics should be 0 by default for BF 4.3 and onwards.

This would be also sort of weird:
Since BF4.0 people are used to turning RPM filtering with just bdir=ON (GUI switch, CLI).
If we change defaults to 0 harmonics we will cause problems cuz people will be thinking they have RPM filters on, but they dont...

KarateBrot:
Okay, I see the problem... In that case I agree. For now (BF 4.3) it's best to turn bidir off at default and make bidir independent in 4.4 and onwards.

limonspb:
@spatzengr 
so filter defaults should have all of this i think:

**set dshot_bidir = OFF**
set rpm_filter_harmonics = 3
set rpm_filter_q = 500
set rpm_filter_min_hz = 100
set rpm_filter_fade_range_hz = 50
set rpm_filter_lpf_hz = 150

spatzengr:
> **set dshot_bidir = OFF**
> set rpm_filter_harmonics = 3
> set rpm_filter_q = 500
> set rpm_filter_min_hz = 100
> set rpm_filter_fade_range_hz = 50
> set rpm_filter_lpf_hz = 150

DONE.

limonspb (approved):

PR #221, opened 2022-03-17
Added CRSF preset for 4.2

MrAlucardDante:
Since all the ELRS presets have the new feedforward, I made a very basic preset that enables Serial RX and set the protocol to CRSF so ELRS (or other RXs that use CRSF) can be configured on 4.2


MrAlucardDante:
The ELRS and TBS presets only work for 4.3 since they are setting the new feedforward values.

I thought that a 4.2 preset could be useful for people that still want to run 4.2 with the new configurator.

But it might be useless, it's up to you to decide.

limonspb (approved):

sugaarK (changes requested):
need to remove this

PR #226, opened 2022-03-22
Gyroflow minimum blackbox settings

Garban5:
Disables all blackblox fields which are not necessary to use [Gyroflow](https://gyroflow.xyz). Only gyro data will be recorded at the lowest logging rate possible. This allows for smaller blackbox files, longer blackbox recordings and lower CPU overhead.

The space gain is substantial. For every 10 seconds of flight, with the minimum configuration the space occupied is approximately 32.5kB and for a typical blackbox configuration (debug mode in Gyro_scaled and a rate of 2kHz) the occupation is 574.7kB
These values are approximate and should be taken as an example. Every drone and every flight are different.

### DIFF
--- a/presets/4.3/other/gyroflow_min_blackbox.txt
+++ b/presets/4.3/other/gyroflow_min_blackbox.txt
@@ -1,0 +1,32 @@
+#$ TITLE: Gyroflow minimum settings
+#$ FIRMWARE_VERSION: 4.3
+#$ CATEGORY: OTHER
+#$ STATUS: EXPERIMENTAL
+#$ KEYWORDS: gyroflow, blackbox, log, DDC, Acrobaix, Garban 
+#$ AUTHOR: Garban
+
+#$ PARSER: MARKED
+
+#$ DESCRIPTION:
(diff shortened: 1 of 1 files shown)


limonspb:
What happens with the previous PR?
Let's not open new PRs for the same thing plz.
I reopened the previous one.

Garban5:
Finally this is the new pull request @sugaarK @limonspb @KarateBrot with the correct file.
sorry for bothering you

limonspb:
> What happens with the previous PR? Let's not open new PRs for the same thing plz. There is always a way to fix commits/branches. I reopened the previous one.

No bother. Just making a new PR is a very last tool :) there is always a way to fix the original. Well always

limonspb (approved):

limonspb:
@KarateBrot  can you have a loot here plz too :)

KarateBrot (approved):

jli:
@Garban5 do you know what's the minimum gyro logging rate that can still get good results with Gyroflow? I'd like to use this preset on a whoop flight controller with just 16MB onboard flash. But the FC also only has a 3.2khz gyro, so 1/16 rate is 200hz.

PR #228, opened 2022-03-23
VTXtable presets compliant with Japanese band plans

ToshihiroMakuuchi:
In Japan, only three waves are available according to the defined band plan. Therefore, it is necessary to prepare a VTXtable preset that complies with the law.


haslinghuis:
@limonspb I smell we need to define country in the feature as configurator setting filtering out geographic rules. We only can enforce using GPS. Geolocation can be spoofed using VPN,

presets/4.3/vtx/Japanese_Bandplans.txt
```diff
@@ -0,0 +1,41 @@
...
+#$ FIRMWARE_VERSION: 4.3
+#$ CATEGORY: VTX
+#$ STATUS: COMMUNITY
+#$ KEYWORDS:  vtx, vtx table, Japan, Japanese, JP, JA, Bandplans
+#$ AUTHOR: ToshihiroMakuuchi
+#$ DESCRIPTION: ----------
+#$ DESCRIPTION: This VTX preset conforms to the Japanese band plans.
+#$ DESCRIPTION: However, it is not a perfect fit for all VTXs. Please use it only as a reference.
+#$ DESCRIPTION:
+#$ DESCRIPTION: -- Amateur Radio License in Japan --
```
limonspb, line 12: It is not clear from the description, if a user can put these VTX channels ONLY when he has Amateur license or everyone can use mentioned channels.

presets/4.3/vtx/Japanese_Bandplans.txt
```diff
@@ -0,0 +1,41 @@
...
+vtxtable channels 8
+vtxtable band 1 BOSCAM_A A FACTORY    0    0    0    0    0    0    0    0
+vtxtable band 2 BOSCAM_B B FACTORY    0    0    0    0    0    0    0    0
+vtxtable band 3 BOSCAM_E E FACTORY 5705    0    0    0    0    0    0    0
+vtxtable band 4 FATSHARK F FACTORY 5740    0    0 5800    0    0    0    0
+vtxtable band 5 RACEBAND R FACTORY    0    0    0    0    0    0    0    0
+vtxtable band 6 LOWRACE L FACTORY    0    0    0    0    0    0    0    0
+#$ OPTION END
+
+#$ OPTION_GROUP END
```
limonspb, line 41: What about power levels in Janap?
Can add option group with powerl levels for different versions of Smart audio and Tramp

presets/4.3/vtx/Japanese_Bandplans.txt
```diff
@@ -0,0 +1,41 @@
...
+#$ STATUS: COMMUNITY
+#$ KEYWORDS:  vtx, vtx table, Japan, Japanese, JP, JA, Bandplans
+#$ AUTHOR: ToshihiroMakuuchi
+#$ DESCRIPTION: ----------
+#$ DESCRIPTION: This VTX preset conforms to the Japanese band plans.
+#$ DESCRIPTION: However, it is not a perfect fit for all VTXs. Please use it only as a reference.
+#$ DESCRIPTION:
+#$ DESCRIPTION: -- Amateur Radio License in Japan --
+#$ DESCRIPTION: Radio communications most commonly available include amateur, civil, and personal radio communications.
+#$ DESCRIPTION: Amateur radio communications are authorized only for those who pass a national license examination or who complete JARD & other official training course and acquire the prescribed radio operator certificate.
```
limonspb, line 14: Could list here the allowed channels for the users that don't want to check the raw CLI text.

presets/4.3/vtx/Freq_change_with_3-6pos_SW.txt
```diff
@@ -0,0 +1,101 @@
...
+#$ CATEGORY: VTX
+#$ STATUS: COMMUNITY
+#$ KEYWORDS:  vtx, vtx table, Japan, Japanese, JP, JA, Jumper-T16, TX16S
+#$ AUTHOR: ToshihiroMakuuchi
+#$ DESCRIPTION: ----------
+#$ DESCRIPTION: Use the 3pos and 6pos switches on the transmitter to set the VTX frequency to change.
+#$ DESCRIPTION: 
+#$ DESCRIPTION: -- Amateur Radio License in Japan --
+#$ DESCRIPTION: Radio communications most commonly available include amateur, civil, and personal radio communications.
+#$ DESCRIPTION: Amateur radio communications are authorized only for those who pass a national license examination or who complete JARD & other official training course and acquire the prescribed radio operator certificate.
```
limonspb, line 13: I think this description is lacking information about how exactly this preset going to work.
As well as option names are not clear. I do not understand what that means for example:
`vtx_power=1&2`
i am pretty sure most user won't understand that too.

limonspb (changes requested):
This is a nice idea.
Left some comments above.

presets/4.3/vtx/Japanese_Bandplans.txt
```diff
@@ -0,0 +1,41 @@
...
+#$ DESCRIPTION: ----------
+
```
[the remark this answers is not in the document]
ToshihiroMakuuchi: Thank you for your response. I will rewrite it for your reference.

presets/4.3/vtx/Japanese_Bandplans.txt
ToshihiroMakuuchi: In Japan, it is available only to holders of amateur radio licenses. This section will be rewritten to be clearer.

presets/4.3/vtx/Japanese_Bandplans.txt
ToshihiroMakuuchi: To use VTX in Japan, the same procedures are required as for opening an amateur radio station.
You must apply to the Ministry of Internal Affairs and Communications for permission to use VTX's maximum output power. Therefore, it is available as in other countries.
However, to avoid interference, it is necessary to use the weakest output possible.

Add the power levels settings for SmartAudio and IRC Tramp.

presets/4.3/vtx/Japanese_Bandplans.txt
ToshihiroMakuuchi: I understand. I will add the permitted channels.
In Japan, seven frequencies are theoretically available, but only three can be used simultaneously due to radio interference. I will add that information as well.

presets/4.3/vtx/Freq_change_with_3-6pos_SW.txt
ToshihiroMakuuchi: I understand. I will rewrite the information that three frequencies at the same time within the scope of the Japanese band plan can be switched immediately using a switch on the transmitter.
We will also reconsider the name used for the option. Please wait a moment.

ToshihiroMakuuchi:
We will close the PR once we re-configure it.

PR #234, opened 2022-04-15
Added Llama Ratebank from Rebase

TehllamaFPV:
Resubmission of old Llama Ratebank with updates.

Added descriptions and review force to ensure users are aware that this can change which profile is selected (but still defaults to the current RateProfile 1 

Actual commit will contain squashed values to update the link to this PR


limonspb (commented):
I do not think that playing with rate profiles is safe in 4.3.
Until we separate TPA from rate profiles (hopefully 4.4) we should not be using profiles in presets.
Otherwise, it will mess up the tune for people. And a messed-up TPA can be pretty dangerous.

TehllamaFPV:
> 

Practically, this just forces the user back to RateProfile 1 (the default), which wouldn't adjust any profile values related to TPA.  
This only alters rate values (and rate types - all legacy) for the 2-6 rate profiles, and only with manually selecting llama rates does it overwrite anything.
The edge case that might affect TPA is if a user has chosen to use a different rate profile # for their own rates, makes a TPA adjustment, then applies this and selects a different rateprofile number - they lose that TPA change (it will be default, or whatever had been previously set).

limonspb:
> 
> Practically, this just forces the user back to RateProfile 1 (the default), which wouldn't adjust any profile values related to TPA. This only alters rate values (and rate types - all legacy) for the 2-6 rate profiles, and only with manually selecting llama rates does it overwrite anything. The edge case that might affect TPA is if a user has chosen to use a different rate profile # for their own rates, makes a TPA adjustment, then applies this and selects a different rateprofile number - they lose that TPA change (it will be default, or whatever had been previously set).

Yup thats what i am talking about, user can't reliably use rate profiles in 4.3 unless he manually sets desired TPA of his tune to all the rate profiles. For that reason, i think that messing with rate profiles is not good, especially in presets. It is just not safe.

TehllamaFPV:
Repackaged as 'OTHER' preset.    Added warning for users to re-check TPA behaviors.

If this version is unsatisfactory, will defer to 4.4

limonspb:
defer to 4.4 as now rate profile switching is not "safe" cuz of TPA

PR #242, opened 2022-05-11
Create ExpressLRS Betalight Passthrough Preset

TehllamaFPV:
Probably the laziest preset ever... but just a quick utility preset that puts the serialrx_halfduplex and serialrx_inverted to default (off) values, and disables telemetry.

### DIFF
--- a/presets/4.3/other/ELRS_Passthrough.txt
+++ b/presets/4.3/other/ELRS_Passthrough.txt
@@ -1,0 +1,16 @@
+#$ TITLE: ExpressLRS Passthrough Betaflight Settings
+#$ FIRMWARE_VERSION: 3.5
+#$ FIRMWARE_VERSION: 4.2
+#$ FIRMWARE_VERSION: 4.3
+#$ STATUS: OFFICIAL
+#$ CATEGORY: OTHER
+#$ KEYWORDS: ELRS, Passthrough, Configure, utility
+#$ AUTHOR: Daniel Appel / Tehllama
+#$ DESCRIPTION: This is a simplified way of ensuring that Betaflight is correctly configured to enable ExpressLRS to perform receiver flashing via Passthrough.
+#$ DESCRIPTION: Backing up flight controller settings is strongly recommended. After completing serial passthrough flashing, configuration should be restored from the backup file.
+#$ WARNING: Please back up flight controller settings using the "Save Backup" button.
+#$ DISCUSSION: https://github.com/betaflight/firmware-presets/pull/242
+
+feature -TELEMETRY
+set serialrx_inverted = off
+set serialrx_halfduplex = off


presets/4.3/other/ELRS_Passthrough.txt
```diff
@@ -0,0 +1,14 @@
+#$ TITLE: ExpressLRS Passthrough Betaflight Settings
+#$ FIRMWARE_VERSION: 3.5
```
limonspb, line 2: are you sure it works with 3.5? why not 4.0 or 4.1?

presets/4.3/other/ELRS_Passthrough.txt
```diff
@@ -0,0 +1,14 @@
...
+#$ FIRMWARE_VERSION: 4.2
+#$ FIRMWARE_VERSION: 4.3
+#$ STATUS: OFFICIAL
+#$ CATEGORY: OTHER
+#$ KEYWORDS: ELRS Betaflight Passthrough Configure
```
limonspb, line 7: keywords divided by comma

presets/4.3/other/ELRS_Passthrough.txt
```diff
@@ -0,0 +1,14 @@
...
+#$ AUTHOR: Daniel Appel / Tehllama
+#$ DESCRIPTION: This is just a simplified way of ensuring that Betaflight is correctly configured to enable ExpressLRS to perform receiver flashing via Passthrough
```
limonspb, line 9: I am not familliar with ERLS at all, so i could be wrong here:
any words of how a person would revert these settings after he is done with passthrough? Or a person does not need to revert them and can keep flying after he applies them?

limonspb (changes requested):

presets/4.3/other/ELRS_Passthrough.txt
TehllamaFPV: I don't have anything on 4.0 or 4.1 to test.  Not a great reason, but I can't see why it wouldn't.

presets/4.3/other/ELRS_Passthrough.txt
TehllamaFPV: Correcting.

presets/4.3/other/ELRS_Passthrough.txt
TehllamaFPV: The reversion is realistically the problem... and why this is potentially a useless preset.
This would allow a user to perform the Betaflight Passthrough method of reflashing an ExpressLRS receiver... but does nothing to restore the previous setting.

presets/4.3/other/ELRS_Passthrough.txt
limonspb: what about 3.5? Does it need to be there/does it works there?

TehllamaFPV: I have tested with 3.5 - it also works with that.  Does not need to be there, but I have performed that testing.

limonspb (dismissed):

limonspb (approved):

PR #252, opened 2022-06-18
Pb street league defaults

LeadFingers:
This has presets for Street League Spec Drone Racing Quads

Street_League_motor_outputs.txt-  sets the spec throttle cap and other spec needed settings along with allowing pilots to select their motor output limit depending on their specific motors


sugaarK:
Can you squash the comments into one then I’ll have a look

limonspb:
clesed because the author did not want to mess his old branch as he is sharing it with people.

PR #254, opened 2022-06-19
mouseFPV tune and filters for FPVCycle Incisor

rochford77:
aka Prototype 5. Discusstion tag will be added  with a link to this PR shortly. Thank you.


presets/4.3/tune/mouseFPV_FFVCycle_Incisor_PT5_tune_filters.txt
```diff
@@ -0,0 +1,151 @@
...
+#$ OPTION END
+
+#$ OPTION BEGIN (UNCHECKED): Heavy Action Cam (Hero8/9/10)
+set simplified_master_multiplier = 120
+simplified_tuning apply
+#$ OPTION END
+
+
+#$ OPTION_GROUP BEGIN: Filters (Choose One or None)
+
```
limonspb, line 104: i would add two options with dshot 300 and dshot 600 with dshot 300 checked.
ANd add to the working something lile: This tune sets ESC protocol to DSHOT, it is extremely unsafe to even select it, if your ESC does not support DSHOT (can spin the motors without arming)

rochford77: Cool, I broke it out into 2 filter options with 300 and 600 respectively, and added the warning per your suggestion.

limonspb: sorry maybe i wasn't clear. YOu copied the whole filtering section.
Check, for example, how it is done here in this preset:
https://github.com/betaflight/firmware-presets/blob/master/presets/4.3/tune/freedom_spec_460g.txt

rochford77:
@limonspb This has been updated to set DShot as per the freedom_spec_460g. 

I also updated the preset to no  longer include FF_transition or boost.

limonspb:
> @limonspb This has been updated to set DShot as per the freedom_spec_460g.
> 
> I also updated the preset to no longer include FF_transition or boost.

Please don't update the ff_boost things here as it is done in my PR and discussion is there. Might cause merge conflicts

rochford77:
@limonspb This file is not in that PR so far as I can see. This has yet to be merged, this was already a open unmerged PR that has ff_boost and ff_jitter_reduction in the open PR. I simply removed those since this PR is still waiting to be merged. Please re-review.

limonspb:
> @limonspb This file is not in that PR so far as I can see. This has yet to be merged, this was already a open unmerged PR that has ff_boost and ff_jitter_reduction in the open PR. I simply removed those since this PR is still waiting to be merged. Please re-review.

oh my bad, good point

limonspb (changes requested):
Seems like this preset needs some refactoring:

Setting motor protocol in TUNE is not allowed unless TUNE includes defaults for the filters.
Again, please check how it is done in freedom spec TUNE:
https://github.com/betaflight/firmware-presets/blob/master/presets/4.3/tune/freedom_spec_460g.txt

If you want your TUNE to be ONLY TUNE without filters, then you can't set a motor protocol.
If you want your TUNE to have Filters `#$ OPTION`, then you set motor protocol there inside the Filters option.

Also a weird situation around
#$ OPTION BEGIN (CHECKED): Dynamic Idle
Let me explain:
The reason you might want to separate TUNE from FILTERS in your preset is when someone already got some filters setup and he wants to try your PIDs (in simple words). In this case the user just skips the "RPM filters option" and all good.
Except for your dynamic idle option, because it sets `set dshot_bidir = ON`.
When we are in the situation where user has his own filters and he does not want to change them, we don't really know if he uses dshot_bidir  or not. So you are basically messing with his filters outside of the Filters Option.

I hope my explanations make sense lol.

So for the simple TUNEs without filters i recommend don't mess dshot,dshot_bidir etc.
If you want to include Filters in your TUNE, do it like it is done in the freedom spec or Karate. And then you can just easily make Dynamic idle option, easily setup dshot60 or 300 etc.
If you want to have an option for filters - thats more complicated and you need to watch carefully and make sure you don't mess with anything outside of the OPTION.

rochford77:
They make sense but I'm becoming a bit frustrated because I considered most of this when you originally requested a change, and submitted a PR that didn't have some of these problems, but that was rejected as well because it "wasn't specifically what you asked....."

So, originally I submitted this PR without ever touching dshot, as is done in my already accepted AOS 3.5 tune.

 You then suggested I break out dshot into 2 options for 300 and 600 with one of them defaulted, because if I don't set dshot before enabling bi-directional, and they add it later, it may cause issues. Okay, no problem. So, I thought about it, and I didn't want to set a motor protocol unless I had to (if only selecting a tune), so I made 2 full filter options that set dshot accordingly within the respective filter option. which still seems like the correct way to do this. You then said 
 
> sorry maybe i wasn't clear. You copied the whole filtering section twice to have dshot300 and dshot600
> Check, for example, how it is done here in this preset:

..and linked the freedom_spec_460g preset.

The reason I did it this way (dshot inside the filter option) is I only wanted someone to have their motor protocol as an option if they weren't selecting a tune. Doing it the way I did it coupled the dshot setting to a filter, to specifically avoid the problems you currently have with the preset (setting dshot without setting filters). 

But, fine, I'll do it the way the freedom spec preset does it because that is what you asked.I looked at that preset and one of the first things it does is set dshot 600 so that's what I did, but now that isn't right either.... apparently because I'm not importing default filters, but I don't want to import filters if they aren't selecting a filter, that seems silly and unexpected behavior from a user perspective....I feel like the way I had it, with 2 full filter options that set the respective dshot setting inside the option was the correct way to do what you are asking, but that was rejected, and I still don't have any idea why.

That seems like the best approach. To offer 2 filter options, "include rpm filter dshot 300" and "include rpm filter dshot 600" that just set the dshot value in the filters option itself. You never gave an explanation for why that wasn't accepted other than "it's not what you asked". Can I just do it that way? I don't see the problem doing it that way....

Also, with dynamic idle, I copied how the UAVTECH presets do it... Do you have a suggestion on how to better accomplish a dynamic ideal option? How did marks get approved but mine did not? How did my AOS3.5 tune get approved but this does not? It's hard because I'm copying how other accepted tunes but then getting my PRs rejected, but only sometimes, I'm fine with strict rules but when they are inconsistent it's hard to know what I need to do. :(

rochford77:
I am not trying to be argumentative or negative, I just feel like im commiting in circles here, and would prefer not to force people into RPM filters in the tune itself, because people with stock bl_s escs wouldnt be able to run the tune at all, and I dont want to make the filters their own file/preset because they dont offer value outside of this tune. I feel like I mostly had it right 2 commits ago, and still dont have a clear understanding of why that was rejected, just feeling a bit defeated, thats all.

presets/4.3/tune/mouseFPV_FFVCycle_Incisor_PT5_tune_filters.txt
```diff
@@ -0,0 +1,161 @@
...
+#$ DESCRIPTION:
+#$ DESCRIPTION: Fly Like mouseFPV | Recommendations outside of tune:
+#$ DESCRIPTION: ----------
+#$ DESCRIPTION: * Apply mouseFPV Freestyle Rates
+#$ DESCRIPTION: * Set Jitter Reduction (feedforward_jitter_factor) to 14
+#$ DESCRIPTION: * Set Feedforward Boost (feedforward_boost) to 10
+#$ DESCRIPTION: 
+#$ DISCUSSION: https://github.com/betaflight/firmware-presets/pull/254
+#$ WARNING:  Included Filters set motor poles to 14! Filters Require RPM filtering and DShot telemetry! This tune sets ESC protocol to DSHOT, it is extremely unsafe to even select it if your ESC does not support DSHOT (can spin the motors without arming)
+#$ DISCLAIMER: Flash At Your Own Risk! This tune sets ESC protocol to DSHOT!
```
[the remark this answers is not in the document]
rochford77: okay cool, I forgot that those were there, will update

rochford77:
okay just forcing the user to use included the filter into the tune. I don't really like it but its the only way I see to meet the requirements to get this merged without cluttering the filter presets section. I will update the AOS tune to do the same since that is not currently compliant, and I planned on updating the description for that preset anyways. also updated to include the stock dshot warning and removed some redundant set commands.

limonspb:
> I am not trying to be argumentative or negative, I just feel like im commiting in circles here, and would prefer not to force people into RPM filters in the tune itself, because people with stock bl_s escs wouldnt be able to run the tune at all, and I dont want to make the filters their own file/preset because they dont offer value outside of this tune. I feel like I mostly had it right 2 commits ago, and still dont have a clear understanding of why that was rejected, just feeling a bit defeated, thats all.

I understand and i really don't want to make your life harder :)
Presets are the new system and we are learning while making them about all of the corner cases. I am learning too.

There are no doubts some presets got approved while still having some problems that i missed.
But I just checked the UAV tech presets and they all are including filter defaults right away. It allows the author to use all of the filter commands without any options. This is a simple way and i think the most preferable way for all the Authors, make it simple for them and users.

Such presets are still powerful. For example, you can make some "default" filters for your preset that don't require RPM filters but still are OK to use with your tune. And then you can provide (CHECKED or UNCHECKED) option that sets the RPM filers.

Now if you want to make some advanced presets for advanced users... for the users that could potentially leave the filters they have and just apply your TUNE, that's more complicated for everyone. Of course its possible tho. But you need to understand the ideas behind presets, and defaults, to think about all the possible outcomes for the users and make sure they got what they select.

The solution with two filter sections, one with DSHOT300, and the other with DSHOT600, both include Filter defaults is ok. My bad, i actually did not notice that you did not include the filter defaults at the very beginning.

But in this scenario, you can't really add the Dynamic Idle option like you have. We don't have nested options yet.
And this is something to fix also in your previous preset that already got merged.

One of the solutions could be a dynamic idle option like that:
```
#$ OPTION BEGIN (CHECKED): Dynamic Idle
    set dshot_idle_value = 150 - i would remove hthis as well.
    set dyn_idle_min_rpm = 20
    set dyn_idle_p_gain = 45
#$ OPTION END
```
without setting DSHOT protocol.
It means that if user has DSHOT protocol selected, he will have dynamic idle. Otherwise, this option does not do anything. Btw you probably want to remove dshot_idle_value from here as it gets ignored by dynamic idle anyways. ANd if dynamic idle not working, 150 is just too low.

I would also suggest you to make a preset that includes FIlters right away tho. Like karate, UAV tech and freedom spec.
And there you provide "default" filter settings that work for users with old blheli_s.

Think about different users:
User that don't update blheli_s is not an advanced user :). He wants simple solutions without extra clicks and flashing.
He wants to apply your preset that wipes out all of his previous mess and makes things good.
I think this is a typical user with old blheli_s. When he applies your preset, he will get your tune, but you are leaving him with his weird filters, whatever they are.

If user has his filters already setup and he wants to just try your PIDs, sounds like he is an advanced user, and he probably have RPM capable ESCs.

rochford77:
hey thanks for the feedback. Please dont review/merge the latest commit ammend until I comment further in a few min, I was doing work, and haphazardly amended my commit, while you were trying your response, and I have more changes to make. Ill comment in a bit when I am ready for the review. thanks for the help

rochford77:
this is going to have to sit on ice for a few days. I have an idea of filters without RPM filtering, but they need to be tested first.

limonspb:
> this is going to have to sit on ice for a few days. I have an idea of filters without RPM filtering, but they need to be tested first.

No rush. You can mark your Par as draft and then just make it regular PR back when you think it's ready for review.

Please don't think I asking for too much from you. Can ask @sugaarK how many times he had to fix his PRs lol. But at the end the result is real nice :)

rochford77:
> No rush. You can mark your PR as draft and then just make it regular PR back when you think it's ready for review.
>

Okay will do

> Please don't think I asking for too much from you. Can ask @sugaarK how many times he had to fix his PRs lol. But at the end the result is real nice :)

No it's no problem. You are great, and I understand the reasoning behind the changes.  The small miscommunication about a previous commit had me feeling frustrated and feeling like "I can't win". Hopefully I wasn't too testy, I was trying to express what had me feeling frustrated without sounding like a jerk. I appreciate your time and your help greatly. :)

sugaarK:
> Please don't think I asking for too much from you. Can ask @sugaarK how many times he had to fix his PRs lol. But at the end the result is real nice :)

😏

rochford77:
I think I have most of the problems solved, open to other ideas though. 

So the general approach is to make this work for anyone. As such, we load the default filters and pids off the bat, and do not enable dshot. We then set pids and filters using the multi dynamic notch rather than RPM filtering. This way, bl_s users can still flash the tune, and we can offer dshot/rpmF options:

There are dshot 300 and dshot 600 rpm filtering options. the 600 is chosen by default. Inside both of these presets, we re-load default filters, and enable the appropriate dshot setting, as well as bi-directional, then apply the proper filters. default warnings for dshot are included in this preset.

also, per @limonspb suggestion, dynamic idle doesnt enable bi directional anymore. so, it will still set the dyn_idle_min_rpm value which wont matter if they dont have bi-directional.

I think that covers all of our issues in a way that wont break any configs. removing this from a draft state and opening back up for review.

presets/4.3/tune/mouseFPV_FFVCycle_Incisor_PT5_tune_filters.txt
```diff
@@ -0,0 +1,242 @@
...
+#$ DESCRIPTION: * **AUW:** 660-670g
+#$ DESCRIPTION:
+#$ DESCRIPTION: ## Fly Like mouseFPV | Recommendations Outside of Tune:
+#$ DESCRIPTION: * Apply mouseFPV Freestyle Rates
+#$ DESCRIPTION: * Use 250hz radio link if possible
+#$ DESCRIPTION: * **Set Jitter Reduction (feedforward_jitter_factor) to 14**
+#$ DESCRIPTION: * Set Feedforward Boost (feedforward_boost) to 10
+#$ DESCRIPTION:
+#$ DISCUSSION: https://github.com/betaflight/firmware-presets/pull/254
+#$ WARNING: If You Choose To Include Filters Please See The Following:
```
limonspb, line 45: If You Choose To Include **RPM** Filters Please See The Following:
...because some filters are included by default

presets/4.3/tune/mouseFPV_FFVCycle_Incisor_PT5_tune_filters.txt
```diff
@@ -0,0 +1,242 @@
...
+# -- Filters for non bi-directional setups as a base--
+
+
+# -- Gyro lowpass filters --
+# -- No Gyro Lowpass
+set gyro_lpf1_static_hz = 0
+set gyro_lpf1_dyn_min_hz = 0
+set gyro_lpf1_dyn_max_hz = 0
+set gyro_lpf2_static_hz = 700
+set simplified_gyro_filter = ON
```
limonspb, line 93: Idk why some people like to add the exact values and then add sliders..
But to make it easier to read the preset code, you can remove all the direct PID/Filter values affected by sliders and just use `simplified_tuning appy`

presets/4.3/tune/mouseFPV_FFVCycle_Incisor_PT5_tune_filters.txt
```diff
@@ -0,0 +1,242 @@
...
+
+# -- Yaw lowpass --
+set yaw_lowpass_hz = 100
+
+#$ OPTION END
+
+#$ OPTION_GROUP END
+
+
+#$ OPTION_GROUP BEGIN: General Options (Choose Many or None)
```
limonspb, line 222: **Extra** options?
General are probably filters, these are just extra

presets/4.3/tune/mouseFPV_FFVCycle_Incisor_PT5_tune_filters.txt
```diff
@@ -0,0 +1,242 @@
+#$ TITLE:  FPVCycle Incisor/Prototype 5 6s Tune | mouseFPV 
+#$ FIRMWARE_VERSION: 4.3
+#$ CATEGORY: TUNE
+#$ STATUS: EXPERIMENTAL
+#$ KEYWORDS: FPVCycle
```
limonspb, line 5: Configurator user keywords for string search when user types something in the search bar (besides title and author name)
So adding more keywords would allow users to find your preset by "5inch" or something like that:
5", 5 inch, 5in, 6S, freestyle, GoPro

limonspb (commented):
Looks good.
I added some comments, but they're optional to fix/change.

rochford77:
Cool, I updated the keywords and changed the verbiage to "additional options" rather than "extra or general". Should be good to rock and roll.

sugaarK:
Just so you know @rochford77  in 4.4 we will have rpm and dynamic idle on as default as I’m making it my mission to have all manufacturers flashing bluejay to blheliS ecs in house

rochford77:
> Just so you know @rochford77 in 4.4 we will have rpm and dynamic idle on as default as I’m making it my mission to have all manufacturers flashing bluejay to blheliS ecs in house

that will be super useful! I still like this approach for now, but I will for sure update/port the presets to a 4.4 version that updates accordingly when the time comes :-)

rochford77:
> > > Just so you know @rochford77 in 4.4 we will have rpm and dynamic idle on as default as I’m making it my mission to have all manufacturers flashing bluejay to blheliS ecs in house
> > 
> > that will be super useful! I still like this approach for now, but I will for sure update/port the presets to a 4.4 version that updates accordingly when the time comes :-)
> 
> Just double checking if you missed some of the comment above by accident or not.

Ah, I think I did, you had some verbiage suggested changes I missed, only saw the bottom 2, my bad! I'll get those updated.

presets/4.3/tune/mouseFPV_FFVCycle_Incisor_PT5_tune_filters.txt
rochford77: Basically, I'm making sure gyro lpf1 is off and lpf2 is on. If I don't set the values first, the slider will leave both the dynamic LPF1 and the static lpf2 enabled and the slider will operate on both. Zeroing out LPF1 seems to flip the UI toggle for LPF1 off, and then the slider just operated the static LPF2.

I suppose the set gyro_lpf2_static_hz = 700 is redundant but the stuff on lpf1 needs to stay, unless there is a command to flip the lpf1 "toggle" in the UI off so that the slider only manipulates the lpf2? Does such a command exist?

presets/4.3/tune/mouseFPV_FFVCycle_Incisor_PT5_tune_filters.txt
rochford77: Will update

PR #256, opened 2022-06-20
added Street League preset tune and motor output limit

LeadFingers:
adds a Street League spec Motor output and throttle limit setting to the "other" preset
adds a LeadFingers PidToolbox tune for Street League quads.

### DIFF
--- a/presets/4.3/tune/Street_League_LeadFingers.txt
+++ b/presets/4.3/tune/Street_League_LeadFingers.txt
@@ -1,0 +1,58 @@
+#$ TITLE: Street League LeadFingers PIDtoolbox tune
+#$ FIRMWARE_VERSION: 4.3
+#$ CATEGORY: TUNE
+#$ STATUS: EXPERIMENTAL
+#$ KEYWORDS: Street League, StreetLeague, Street, League, LeadFingers, PIDtoolbox, 7"
+#$ AUTHOR: LeadFingers
+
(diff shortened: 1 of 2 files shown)


limonspb (approved):

presets/4.3/tune/Street_League_LeadFingers.txt
```diff
@@ -0,0 +1,58 @@
...
+set gyro_lpf1_static_hz = 0
+set gyro_lpf2_static_hz = 200
+set dyn_notch_q = 400
+set dyn_notch_min_hz = 110
+set dyn_notch_max_hz = 400
+set gyro_lpf1_dyn_min_hz = 0
+set gyro_lpf1_dyn_max_hz = 0
+set simplified_gyro_filter = OFF
+set rpm_filter_harmonics = 1
+set rpm_filter_q = 1000
```
sugaarK, line 30: High rpm Q is somthing we discovered in developing the cross fade code to be less than desirable. The anti phase signal for your rpm takes a longer to dissipated than it does to cancel out the noise peek creating a kinda ghost signal. Better to increase the cross fade range to say 100hz than to push the rpm Q

LeadFingers: Thank you for the suggestion!  After discussing this with Brian who helped me with the tune I think I'm going to leave it as is for now. I'll have some time to test new filters next month and we can make an update if the black box logs look good.

PR #41, opened 2021-10-10
Update karate_race_tune_preset.txt

sugaarK:
tried to improve the readability of the description and added relevant pilot names to the key words

PR #258, opened 2022-06-21
Karate Race 6S polishing

limonspb:
1. Added experimental BOSH option
2. Reduced AG gain to 3.0
3. Adding corresponding rc_smoothing to the RC_LINK options
4. Replaced limon's spicy D gain with Spicy TPA
5. Added options_group for RC_LINKS for better look
6. Added indentation to the code

Original Karate PR:
https://github.com/betaflight/firmware-presets/pull/41

### DIFF
--- a/presets/4.3/tune/karate_race.txt
+++ b/presets/4.3/tune/karate_race.txt
@@ -4,27 +4,45 @@
 #$ STATUS: OFFICIAL
 #$ KEYWORDS: karate, 6S, race, world champion, 5 inch, 5", sugarK, limon, ctzsnooze, karateBrot
 #$ AUTHOR: sugarK
-#$ DESCRIPTION: This 6s racing tune was developed with the help of @ctzsnooze around @karatebrot’s RPM crossfade code allowing a pretty aggressive filter array which is the karate array.
-#$ DESCRIPTION: This tune works well on any quality airframes and has been flown on the Viper racer (where it was developed ) BMS Js1, 533 switchback, HGLRC wind and of course @limons Open racer.
-#$ DESCRIPTION: It should also work on other quality racing frames but use at your own risk as it is an aggressive tune. This tune so far has won a good number of events including the 2021 IO.
-#$ DESCRIPTION:
-#$ DESCRIPTION: Things to note: YOU HAVE TO USE RPM FILTERING WITH THIS TUNE, failure to do so might result in fire and regarding Dshot600,  if your setup has errors in the motor tab using bidirectional Dshot then change to 8k/4k and dshot300.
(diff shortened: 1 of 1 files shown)


sugaarK:
we need a logo let Me speak to @KarateBrot

limonspb:
> we need a logo let Me speak to @KarateBrot

logo in the next PR lol

sugaarK:
lols just the logo it doesn't need the karate!!! needs a roundel though cause it'll look funny in dark mode


I think that's perfect though

limonspb:
> lols just the logo it doesn't need the karate!!! needs a roundel though cause it'll look funny in dark mode
> 
> I think that's perfect though

lets review this PR without logo plz.

sugaarK (commented):
oops

presets/4.3/tune/karate_race.txt
```diff
@@ -26,6 +26,11 @@ set gyro_lpf1_dyn_min_hz = 500
 set gyro_lpf1_dyn_max_hz = 1000
 set simplified_gyro_filter = OFF
 
+#$ OPTION BEGIN (UNCHECKED): Bosh (BMI) gyro - experimantal
```
sugaarK, line 29: shouldn't this have the OSR4 cli command just to be sure... also ive had a number of people contact me and say they couldn't get the Bosch flying good so my solution was to kick the fixed stage2 gyro back to 250hz..

sugaarK: I don't think all these Fcs are equal.. its the usual issue im seeing though not enough hardware filtering

sugaarK (approved):

presets/4.3/tune/karate_race.txt
limonspb: > I don't think all these Fcs are equal.. its the usual issue im seeing though not enough hardware filtering

so there are FCs with BMI gyro + OSR4 mode that requires MORE filtering that MPU6000? That's pretty wild for OSR4 mode.

limonspb:
added logo as per @KarateBrot but moved it to the right. Square logos accupy too much room in the middle shifting the description too low :)

image

KarateBrot (approved):

dezignhero:
@sugaarK @limonspb Following up from a discussion on Slack.  Would it be possible to add a "safe" range for the Dynamic Idle setting?  Since it's written as 2000kv I assume I need to have 2000kv exactly, but I've been told it's actually usable to for 1800kv-2200kv range which is quite broad.  Therefore could we change `#dynamic idle for 6s 2000kv` to `#dynamic idle for 6s 1800-2200kv` ?

PR #262, opened 2022-06-22
KarateTune: Stylize Description

KarateBrot:
Follow-up to https://github.com/betaflight/firmware-presets/pull/258

Only minor markdown styling and an option group for sugarK's rates to visually separate it a bit more from the options above.


sugaarK (approved):
amazing

PR #264, opened 2022-06-24
Removed feedforward_jitter_factor and feedforward_averaging froom TUNE

limonspb:
Both `feedforward_jitter_factor` and `feedforward_averaging` are parts of the RC_LINK category
so TUNEs don't need and can't set these values. Unless inside the RC_LINK options.
The final CLI outcome must be the same regardless of the order that user applies presets (RC_LINK + TUNE == TUNE + RC_LINK)

So i removed `feedforward_jitter_factor` from tune/defaults.txt
removed `feedforward_jitter_factor` and `feedforward_averaging` from the tunes
replaced old RC_LINK options in the tunes with the proper `#$ INCLUDE rc_link/....`

Please check the changes with care.

### DIFF
--- a/presets/4.3/other/Tehllama_5in_2350-2750KV-TexasSpec_3S-4S-5S_Race_TUNE_OTHER.txt
+++ b/presets/4.3/other/Tehllama_5in_2350-2750KV-TexasSpec_3S-4S-5S_Race_TUNE_OTHER.txt
@@ -10,7 +10,7 @@
 #$ DESCRIPTION: Strongly recommend a full chip erase reflash if alternative tunes are desired.
 #$ DESCRIPTION: Extensive testing has been done across 57+ builds, however not every craft will run best on this tune.
 #$ DESCRIPTION: This tune will auto-select profiles based on battery voltage at plugin by default, but only for 3S, 4S and 5S packs. 
-#$ WARNING: Use at your own risk.  NO 6-cell tune or throttle limit is provided with this tune.  Plugging in a 6S pack with 3S profile active will likely cause SERIOUS damage.  
+#$ WARNING: Use at your own risk.  NO 6-cell tune or throttle limit is provided with this tune.  Plugging in a 6S pack with 3S profile active will likely cause SERIOUS damage.
 #$ DISCUSSION: https://github.com/betaflight/firmware-presets/pull/236
 
 # FORCE_OPTIONS_REVIEW: TRUE
@@ -44,7 +44,6 @@ set d_min_roll = 28
 set d_min_pitch = 35
 set d_max_advance = 0
 set auto_profile_cell_count = 4
-set feedforward_jitter_factor = 12
 set feedforward_boost = 12
 set throttle_boost = 12
 
@@ -82,7 +81,6 @@ set d_min_pitch = 26
 set d_max_advance = 0
 set motor_output_limit = 92
 set auto_profile_cell_count = 5
-set feedforward_jitter_factor = 12
 set feedforward_boost = 8
 set throttle_boost = 5
 set feedforward_max_rate_limit = 103
@@ -121,7 +119,6 @@ set d_min_pitch = 35
 set d_max_advance = 0
 set motor_output_limit = 81
 set auto_profile_cell_count = 3
-set feedforward_jitter_factor = 12
 set feedforward_boost = 18
 set throttle_boost = 18
 set feedforward_max_rate_limit = 103
--- a/presets/4.3/tune/SupaflyFPV_Freestyle_5_Inch_EasyTune.txt
+++ b/presets/4.3/tune/SupaflyFPV_Freestyle_5_Inch_EasyTune.txt
@@ -47,10 +47,6 @@ set dterm_lpf2_static_hz = 140
 
 set anti_gravity_gain = 4000
 
-# -- Feedforward jitter reduction --
-
-set feedforward_jitter_factor = 12
-
 #$ OPTION BEGIN (CHECKED): RPM filtering
 
 set motor_pwm_protocol = DSHOT600
--- a/presets/4.3/tune/SupaflyFPV_Freestyle_6_and_EasyTune.txt
+++ b/presets/4.3/tune/SupaflyFPV_Freestyle_6_and_EasyTune.txt
@@ -47,10 +47,6 @@ set dterm_lpf2_static_hz = 140
 
 set anti_gravity_gain = 4000
 
-# -- Feedforward jitter reduction --
-
-set feedforward_jitter_factor = 12
-
 #$ OPTION BEGIN (CHECKED): RPM filtering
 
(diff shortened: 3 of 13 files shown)


limonspb:
[rochford77](https://github.com/rochford77)
[FPVian](https://github.com/FPVian)

SupaflyFPV:
Jitter factor has a dual purpose - it was designed to deal with jitter but it has now taken over from Feedforward transition. It is used as a tuning solution to soften the action of feedforward so I think it should be maintained also for a TUNE parameter as well as RC LINK...

limonspb:
> Jitter factor has a dual purpose - it was designed to deal with jitter but it has also now taken over from Feedforward transition. It is used as a tuning solution to soften the action of feedforward and is an important tuning parameter now - so I think it should be maintained also for a TUNE parameter as well as RC LINK...is there a reason it cannot exist in both categories?

Existing in both categories means the final value for the user is unknown. Today user applies TUNE first, then RC_link, tomorrow for the second quad he applies RC_LINK first and the TUNE.

SupaflyFPV:
> > Jitter factor has a dual purpose - it was designed to deal with jitter but it has also now taken over from Feedforward transition. It is used as a tuning solution to soften the action of feedforward and is an important tuning parameter now - so I think it should be maintained also for a TUNE parameter as well as RC LINK...is there a reason it cannot exist in both categories?
> 
> Existing in both categories means the final value for the user is unknown. Today user applies TUNE first, then RC_link, tomorrow for the second quad he applies RC_LINK first and the TUNE.

Ok I see the issue...

Can you give me some time to explore the RC Link categorisation/organisation because currently I would prefer to keep:

Jitter factor
FF Boost

...in the TUNE categories seeing as it influences the feel and behaviour which is part of the 'seasoning' for the tunes...

I see some 'feel based' categories such as 'Freestyle, Cinematic' etc are existing in the RC LINK settings...

For instance there is an argument for aggressive freestyle to have low RC Smoothing (and lower RC latency like racer) but high jitter factor (less aggressive FF) - not currently an option....

@ctzsnooze is away I think unfortunately....

limonspb:
> > > Jitter factor has a dual purpose - it was designed to deal with jitter but it has also now taken over from Feedforward transition. It is used as a tuning solution to soften the action of feedforward and is an important tuning parameter now - so I think it should be maintained also for a TUNE parameter as well as RC LINK...is there a reason it cannot exist in both categories?
> > 
> > 
> > Existing in both categories means the final value for the user is unknown. Today user applies TUNE first, then RC_link, tomorrow for the second quad he applies RC_LINK first and the TUNE.
> 
> Ok I see the issue...
> 
> Can you give me some time to explore the RC Link categorisation/organisation because currently I would prefer to keep:
> 
> Jitter factor FF Boost
> 
> ...in the TUNE categories (as well) seeing as it influences the feel and behaviour which is part of the 'seasoning' for the tunes...
> 
> I see some 'feel based' categories such as 'Freestyle, Cinematic' etc are existing in the RC LINK settings...
> 
> ...but there is nothing for the scenario of aggressive freestyle - to have low RC Smoothing (and lower RC latency like racer) but high jitter factor (less aggressive FF)....
> 
> @ctzsnooze is away I think unfortunately....

ff_boost should be removed from TUNE too to guarantee the outcome, well, in the next PR.
we have 5 RC_LINK groups:
race, freestyle, HD_freestyle, cinematic, ultra_cinematic. This should be enough i think.
But we can make a new rc_link "aggressive freestyle" if we really need it. Then also "smooth_racing" and then "soft_freestyle" and then "aggressive_cinematic" :))

SupaflyFPV:
to add to the previous discussion, these mentioned duplicated parameter settings in dispute are also in TUNE presets:

https://github.com/betaflight/firmware-presets/blob/master/presets/4.3/tune/ctzsnooze_5in_4S_race.txt

https://github.com/betaflight/firmware-presets/blob/master/presets/4.3/tune/karate_freestyle_5_inch.txt

https://github.com/betaflight/firmware-presets/blob/master/presets/4.3/tune/karate_race.txt

https://github.com/betaflight/firmware-presets/blob/master/presets/4.3/tune/tiny_karate.txt

https://github.com/betaflight/firmware-presets/blob/master/presets/4.3/tune/fpvian_basher_freestyle.txt

https://github.com/betaflight/firmware-presets/blob/master/presets/4.3/tune/mouseFPV_AOS_35_tune_filters.txt

https://github.com/betaflight/firmware-presets/blob/master/presets/4.3/tune/whoop_justice.txt

....

Karate Race also does separate RC settings as option - which could be a good potential solution....

limonspb:
> to add to the previous discussion, these mentioned duplicated parameter settings in dispute are also in TUNE presets:
> 
> https://github.com/betaflight/firmware-presets/blob/master/presets/4.3/tune/ctzsnooze_5in_4S_race.txt
> 
> https://github.com/betaflight/firmware-presets/blob/master/presets/4.3/tune/karate_freestyle_5_inch.txt
> 
> https://github.com/betaflight/firmware-presets/blob/master/presets/4.3/tune/karate_race.txt
> 
> https://github.com/betaflight/firmware-presets/blob/master/presets/4.3/tune/tiny_karate.txt
> 
> https://github.com/betaflight/firmware-presets/blob/master/presets/4.3/tune/fpvian_basher_freestyle.txt
> 
> https://github.com/betaflight/firmware-presets/blob/master/presets/4.3/tune/mouseFPV_AOS_35_tune_filters.txt
> 
> https://github.com/betaflight/firmware-presets/blob/master/presets/4.3/tune/whoop_justice.txt
> 
> ....
> 
> Karate Race also does separate RC settings as option - which could be a good potential solution....

Yes in this PR I fixed them all, and replaced some old RC link options with the new ones (in the past rc_link didn't include rc_smoothig)

SupaflyFPV:
Ok......as far as the PR is concerned I'm fine for the changes laid out (removing jitter factor in my presets). Maybe the others have some comments. As mentioned as far as the preset functionality is concerned I see the issue requires fixing.

In my next tune update I will consider the inclusion of RC Link parameters as options, or with CTZ we can discuss the RC Link categories.

sugaarK (approved):
I have to agree with @limonspb its very much a  flying style/radio link/rc smoothing function. Until we have the settings auto configurator to suit the rc link speed it’s very link dependant

KarateBrot (approved):

SupaflyFPV:
ok. Maybe the positioning of this parameter has been changed since I last was involved. The crossover nature Chris explained to me is outlined in the tuning notes:

https://github.com/betaflight/betaflight/wiki/4.3-Tuning-Notes#Feedforward-jitter-reduction

FPVian:
Thanks for tagging me! In the future it would be nice to bring these back to tuning presets, but the way things are today, I agree it makes more sense to keep them exclusive to RC_LINK.

PR #265, opened 2022-06-25
Removed ff_boost from the tune presets

limonspb:
Continuation of this PR:
https://github.com/betaflight/firmware-presets/pull/264

`feedforward_boost` is a part of RC_LINK so it varies depending on race/freestyle/cinematic and RC_LINK.
So TUNE presets should not be setting/resetting it except inside the complimentary RC_LINK options. And TUNE authors should rely on RC_LINK category to set this parameter.


limonspb:
@SupaflyFPV @TehllamaFPV @ctzsnooze  your presets are affected with this PR

sugaarK (approved):
I approve

rochford77:
Not like it matters but....I feel like they have an impact on setpoint tracking and thus DO belong in tunes, and smoothing and averaging belong in the radio link presets, but I digress, you all know better than me. Just wanted to say my peace. 

That said... A small point of consideration as things move forward to make it more clear to us what belongs where: Anything on the PID profile settings page in BFC should be open to tune presets. If a setting isn't applicable for a tune preset, it belongs somewhere else in BFC. IE, if Jitter Reduction is a link setting and not a tune setting, it belongs on the Receiver tab in BFC, ect. 

It makes the most sense for the preset categories to fully align with pages/tabs in BFC. Presets should not need to span tabs/pages in BFC. IMO this also extends to things like TPA being a rate-profile setting and not a pid profile setting but I imagine that is a can of worms that goes back pretty far, maybe its all better left alone, but in my opinion BFC tabs/pages and respective presets should be fully aligned some time in the future for clarity to users and preset contributors. 

-mouse

limonspb:
> Not like it matters but....I feel like they have an impact on setpoint tracking and thus DO belong in tunes, and smoothing and averaging belong in the radio link presets, but I digress, you all know better than me. Just wanted to say my peace.
> 
> That said... A small point of consideration as things move forward to make it more clear to us what belongs where: Anything on the PID profile settings page in BFC should be open to tune presets. If a setting isn't applicable for a tune preset, it belongs somewhere else in BFC. IE, if Jitter Reduction is a link setting and not a tune setting, it belongs on the Receiver tab in BFC, ect.
> 
> It makes the most sense for the preset categories to fully align with pages/tabs in BFC. Presets should not need to span tabs/pages in BFC. IMO this also extends to things like TPA being a rate-profile setting and not a pid profile setting but I imagine that is a can of worms that goes back pretty far, maybe its all better left alone, but in my opinion BFC tabs/pages and respective presets should be fully aligned some time in the future for clarity to users and preset contributors.
> 
> -mouse

Yes I agree that at the end it should align with the TABs, and especially TPA in rates is the most annoying lol.

As for the ff_boost being in the TUNE, I have no idea lol. I only know I can't be in both TUNE and RC_LINK. For the rest I'm following @ctzsnooze and @sugaarK as they are the masters.

KarateBrot (approved):

SupaflyFPV:
> Not like it matters but....I feel like they have an impact on setpoint tracking and thus DO belong in tunes, and smoothing and averaging belong in the radio link presets, but I digress, you all know better than me. Just wanted to say my peace.
> 
> That said... A small point of consideration as things move forward to make it more clear to us what belongs where: Anything on the PID profile settings page in BFC should be open to tune presets. If a setting isn't applicable for a tune preset, it belongs somewhere else in BFC. IE, if Jitter Reduction is a link setting and not a tune setting, it belongs on the Receiver tab in BFC, ect.
> 
> It makes the most sense for the preset categories to fully align with pages/tabs in BFC. Presets should not need to span tabs/pages in BFC. IMO this also extends to things like TPA being a rate-profile setting and not a pid profile setting but I imagine that is a can of worms that goes back pretty far, maybe its all better left alone, but in my opinion BFC tabs/pages and respective presets should be fully aligned some time in the future for clarity to users and preset contributors.
> 
> -mouse

the problem is that (and I didfn't know this until just recently) the jitter factor is heavily dependent on link rate so if we set it for tuning on one link rate - it will be different at another...so we need to work on another solution to make it consistent which I think is in progress...with regard to tune I agree with you, and there are some other FF parameters which cross over, however we can offer options (unchecked) in our tunes so that kind of sorts it out for now.

TehllamaFPV:
I'm probably an edge case user, but I actually do implement the feedforward_boost behavior as one of the multi-voltage tuning levers in order to try and make the feedforward response as consistent as possible. I don't really have a big pile of A/B test data to back it up, but I consider it to be functionally a tune performance parameter more than an RC_Link tuning one, but I would also be fine with this change overall since most of those tune setups are also actively manipulating TPA and trying to do other things simultaneously (like apply multi-voltage filter packages).
I do concur that the ff_jitter and smoothing are very much RC_Link driven parameters, but I haven't really investigated any significant links between those and the boost parameter, but I do actually alter the boost parameter across a lot of tunes (but it's consistently conflated with higher FF gains because I'm changing master slider per the Supafly calculator in most cases).

For the Configurator side of things, I'd like the tabs and presets to all fully align.

For the Presets side of things, I actually have a mildly absurd feature request (in Limon's copious spare time, lol)... it would be introducing a #$ PRECEDENCE (HIGH//LOW//DEFAULT) type behavior where a parameter can be shared across multiple preset classes, and be set as a high priority, low (middle) priority, or if a character regex pulls 'default' from that file uses a 'lowest' priority setting to where a parameter can be set, but would be overwritten.

PR #269, opened 2022-06-29
Creating Preset for Toothpick 3 on 3s

rochford77:
This is a preset for 3" toothpick builds. 

Offers SDFT filters in the main preset, and has options for RPMf on f4 or f7, and those set dshot accordingly. We also have a dyn_idle option that does not set dshot or bi-directional, so it will fail silently/gracefully if those aren't on (the value just wont do anything. 

The TP3 is a fully symmetrical machine, true x with even mass distribution, other than the lipo. The lipo in this quad is side mount, so it goes against BF standard practice and actually has higher gains on roll than pitch because of this. However, similar frames that mount the lipo front to back, or an improper build of the tp3 with the lipo going front to back would need the pid values "flipped" about pitch and roll, such that gains were higher on pitch. There is an option that flips the pid values (via the sliders, but the sliders are not really "flipped") for pitch and roll to accomplish this.

will update description with this PR in a moment.

### DIFF
--- a/presets/4.3/tune/mouseFPV_toothpick3_3s.txt
+++ b/presets/4.3/tune/mouseFPV_toothpick3_3s.txt
@@ -1,0 +1,225 @@
+#$ TITLE:  FPVCycle Toothpick 3 3s | mouseFPV
+#$ FIRMWARE_VERSION: 4.3
+#$ CATEGORY: TUNE
+#$ STATUS: EXPERIMENTAL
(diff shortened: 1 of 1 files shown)


limonspb (approved):
perfect from the first try seems like!

I started doing indentations for the options and option groups content like:
```
#$ OPTION
    set ...
    set ...
    set ...
#$ OPTION END
```

but it is not enforced, maybe yet :)

PR #270, opened 2022-06-30
remove spicy TPA options cuz it's not actually spicy

limonspb:
Thanks to @ctzsnooze for pointing out that higher TPA values makes things less spicy :)
Removing this shame asap

Removed a few trailing spaces also


sugaarK (approved):
Lols

PR #271, opened 2022-06-30
Presets update

SupaflyFPV:
Hi Dudes

Got updates for my presets and new formatting. 

I've tested the formatting and presets - all good...however - I may have messed up the squashing and merging...so its probably all in a pickle...thought may as well start here tho...

### DIFF
--- a/presets/4.3/tune/SupaflyFPV_Freestyle_5_Inch_EasyTune.txt
+++ b/presets/4.3/tune/SupaflyFPV_Freestyle_5_Inch_EasyTune.txt
@@ -4,33 +4,66 @@
 #$ STATUS: OFFICIAL
 #$ KEYWORDS: Supafly, SupaflyFPV, 4S, 6S, Freestyle, 5 inch, 5"
 #$ AUTHOR: SupaflyFPV
-#$ DESCRIPTION: SupaflyFPV Community Freestyle Preset for 5 Inch Builds
+
+#$ PARSER: MARKED
+
+#$ DESCRIPTION: [<img src="https://supaflyfpv.com/wp-content/uploads/2022/06/Logo-Cropped192.svg" width="220px" style="margin-left: auto; margin-right: auto; display: block;"/>](https://supaflyfpv.com/betaflight-setup/)
 #$ DESCRIPTION:
-#$ DESCRIPTION: Select the option for your build with or without HD camera (eg GoPro) as the baseline.
+#$ DESCRIPTION: <br>
 #$ DESCRIPTION:
-#$ DESCRIPTION: Half to one click up or down Master Slider to tweak the tune - in OSD or Configurator.
+#$ DESCRIPTION: ### SupaflyFPV 5 Inch Freestyle Preset
 #$ DESCRIPTION:
-#$ DESCRIPTION: For a Master Slider setting tailored to your Lipo and Motor Kv go here: https://tinyurl.com/SupaflyFPV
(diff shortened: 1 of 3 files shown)


SupaflyFPV:
confused.com about the conflict above...

KarateBrot:
It means that someone else made changes to these files which became part of master. So you are using an outdated version of master. You need to rebase your branch (update your branch to include the latest changes from master) before you can merge your additional changes to the files.

limonspb:
> confused.com about the conflict below...

your branch was 10 commits behind the master branch before you started updating your presets.
It is good to always rebase on master before any new changes.

On the side note:
` OPTION BEGIN (UNCHECKED): Supafly Fast&Smooth RC Link settings `
please make sure you include rc link defaults into this option.

Also im am not sure how a single RC link option can work for all rc link speeds 30 - 500hz and protocols.
Pretty sure you had some RC link speed in mind when you set these values, plz mention that in the #$ option name.

SupaflyFPV:
> Also i am not sure how a single RC link option can cover/work for all rc link speeds 30hz - 500hz and all protocols. Pretty sure you had some RC link speed in mind when you set these values, plz mention that in the #$ option name.

There is no need to mention link speed because I have used settings which is not far from default, and the RC autosmoothing has inbuilt compensator for RC link speed. The setting is designed to work widely, and I have logged it on the most common link in freestyle (CF 50hz).

limonspb:
> > Also i am not sure how a single RC link option can cover/work for all rc link speeds 30hz - 500hz and all protocols. Pretty sure you had some RC link speed in mind when you set these values, plz mention that in the #$ option name.
> 
> There is no need to mention link speed because I have used settings which is not far from default, and the RC autosmoothing has inbuilt compensator for RC link speed. The setting is designed to work widely, and I have logged it on the most common link in freestyle (CF 50hz).
> 
> EDIT: on consideration I think I'm going to give up this RC_LINK battle and remove settings.

1. It is not about how close defaults to what you have in mind, its about making sure the user gets the same RC link values no matter which RC link mess he had before. Including defaults guarantees the same outcome for the same RC LINK option selection regardless of the user's previous values.

2. If you tuned it for 50Hz link speed please mention that in the RC LINK option. Different RC speeds require different rc_smoothing_auto_factor, feedforward_smooth_factor and many other parameters like feedforward_averaging etc. A mismatch between these parameters and RC_LINK will cause extra feedforward noise/setpoint noise and inconsistency of the "feel" across different RC_LINKs.

SupaflyFPV:
yes even though I think the majority of these links are outliers, Im going to leave it out for this reason

limonspb:
> 

you can call them outliers or not, but they are way different for different RC_LINK speeds and that's how @ctzsnooze is planned then.

limonspb:
> yes even though I think the majority of these links in the case of freestylers are outliers, Im going to leave it out for this reason

This is how feedforward looks like at 500hz ghost on your preset versus proper 500hz freestyle preset.
I'm just saying that we have to mention at least which RC link your options is made for. Cuz it can't be made for all of the speeds and you are just misguiding users:
image
[comparison.zip](https://github.com/betaflight/firmware-presets/files/9024291/comparison.zip)

SupaflyFPV:
its good to see a comparison thanks for that. The setpoint is fine with the auto setting, but the FF thats a real problem and I had not seen it in a log before - in freestyle world its rare to see any RC Link above 150hz.

PR #275, opened 2022-07-07
Create mouseFPV preset for FPVCycle Sonicare

rochford77:
The FPVCycle Sonicare is a new frame from FPVCycle. This frame has a unique wishbone arm design that uses 8 individual arm members to build the 4 arms. This quad has amazing resonance performance and very little low freq noise, meaning we can push the pids pretty far and reduce filtering significantly.

This tune loads the default tune and filters, then applies a non-rpm filtering setup as well as proper pids. There are options for ds300/600 RPM filtering setups that re-load the default filters, apply the proper dshot and filter settings. dshot300 assumes a 4k pidloop will be needed, and thus adds in an lpf2 on the gyro.

There is a dynamic idle option that does not enable dshot, so if things are not properly setup, it will set the value but it wont actually do anything.

### DIFF
--- a/presets/4.3/tune/mouseFPV_FPVCycle_Sonicare_tune_filters.txt
+++ b/presets/4.3/tune/mouseFPV_FPVCycle_Sonicare_tune_filters.txt
@@ -1,0 +1,206 @@
+#$ TITLE:  FPVCycle Sonicare 6s Tune | mouseFPV 
+#$ FIRMWARE_VERSION: 4.3
+#$ CATEGORY: TUNE
+#$ STATUS: EXPERIMENTAL
+#$ KEYWORDS: FPVCycle
+#$ AUTHOR: mouseFPV
+#$ PARSER: MARKED
+#$ DESCRIPTION: <br>
+#$ DESCRIPTION: <img src="https://user-images.githubusercontent.com/19867640/174462482-28bdcfec-1c3a-43db-99d7-50688b92f050.svg" width="100px" style="margin-left: auto; margin-right: auto; display: block;"/>
+#$ DESCRIPTION:
+#$ DESCRIPTION: # Tune for FPVCycle Sonicare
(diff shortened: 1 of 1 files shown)


sugaarK:
where is the link to the frame

rochford77:
> where is the link to the frame

@sugaarK are you just curious, or are you implying I should include a link in the preset? If you are  just curious:

https://fpvcycle.com/products/fpvcycle-sonicare-x5-frame-kit?_pos=2&_sid=2050a8e41&_ss=r

I wasn't sure how "commercial" betaflight wanted to get with including links to products, so I haven't been including links to the frames for my frame specific tunes. Most of them are FPVCycle, so there is only "one shop that carries them", but if it were for a more mainstream frame, I could see someone bitching "Why do betaflight presets point to RaceDayQuads and not GetFPV!?!" ect... But I would be happy to include it in the preset if you think that is appropriate, and I can double back and update my other frame specific tunes to include links to the frame. Maybe we can limit links to the manufacturer pages and not 3rd party store pages?

sugaarK:
It makes sense to. My frame preset has a link to the rotobuilds page for it. If I had a shop page it would be there for sure

KarateBrot:
@rochford77 has a good point imo. If you had your own shop, that seems alright, but if we are talking about third party shops, it seems unfair to link people to one particular shop when there are several who are selling the frame. I feel like we should stay neutral unless the author of a preset is linking to his own website (for example if you're the owner of a frame, selling the frame exclusively in your own shop).

sugaarK:
Good point @KarateBrot

limonspb (approved):

rochford77:
Hey @limonspb I just wanted to say thanks for all your help with this stuff.

limonspb:
> Hey @limonspb I just wanted to say thanks for all your help with this stuff.

Test

PR #287, opened 2022-09-26
Adds MSP option to OpenVTx

JyeSmith:
This PR adds the MSP option to the OpenVTx preset.

Since the normal MSP setup requires 4 user steps (select port, baud, msp-vtx, and enter the CLI to set MSP to half duplex) I have tried to simplify this for the users by added the port selection to the preset.  A single check and everything is auto setup 😃 

There are also minor changes to the KEYWORDS and to uncheck TRAMP as the default.  MSP should be encouraged due to the auto configuration of the VTx tables.  Should the DISCUSSION be updated to also include this PR?

To test, OpenVTx master can be flashed with PIO or a binary downloaded from [here ](https://github.com/OpenVTx/OpenVTx/pull/40#issuecomment-1170598763) and the Configurator used to upload https://openvtx.org/

### DIFF
--- a/presets/4.3/vtx/OpenVTx.txt
+++ b/presets/4.3/vtx/OpenVTx.txt
@@ -1,21 +1,36 @@
 #$ TITLE: OpenVTx
 #$ FIRMWARE_VERSION: 4.2
 #$ FIRMWARE_VERSION: 4.3
+#$ FIRMWARE_VERSION: 4.4
 #$ CATEGORY: VTX
 #$ STATUS: COMMUNITY
-#$ KEYWORDS:  vtx, vtx table, OpenVTx, TRAMP, SA 2.1, E7082VM, OVX300, OVX303, OP400
+#$ KEYWORDS:  vtx, vtx table, OpenVTx, TRAMP, SA 2.1, MSP
 #$ AUTHOR: Jye Smith
+
+#$ PARSER: MARKED
+
+#$ DESCRIPTION: 
+#$ DESCRIPTION: <img src="https://raw.githubusercontent.com/OpenVTx/Openvtx-configurator/master/src/assets/logo.png" width="235px" height="64" style="object-fit: cover; margin-left: auto; margin-right: auto; display: block;"/>
+#$ DESCRIPTION: 
 #$ DESCRIPTION: VTx tables for OpenVTx capable video transmitters.
-#$ DESCRIPTION:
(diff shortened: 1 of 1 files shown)


JyeSmith:
> MSP over VTX is only available on 4.4 ?

Yep, so what will be the best way to hand that?  SA and TRAMP are supported in earlier versions, and I dont want to end up with 2 presets for different versions.

haslinghuis:
Afaik you have to make a preset for 4.2 / 4.3 with SA / TRAMP and a 4.4 preset with MSP (as SA & TRAMP won't be needed anymore in 4.4). @limonspb are there any provisions to make this happen in one preset?

limonspb:
> Afaik you have to make a preset for 4.2 / 4.3 with SA / TRAMP and a 4.4 preset with MSP (as SA & TRAMP won't be needed anymore in 4.4). @limonspb are there any provisions to make this happen in one preset?

@JyeSmith you can name your new MSP group like that:
#$ OPTION_GROUP BEGIN: MSP (Betaflight 4.4+)
Just to notify users that these options won't work in 4.2 and 4.3
Or two separate presets are also fine.

@haslinghuis but that's a good idea for me to think about in the future.

limonspb (changes requested):
@JyeSmith besides adding (Betaflight 4.4+) in the group name i would also suggest to add
`#$ FORCE_OPTIONS_REVIEW`, usually goes somewhere before or after #$DISCUSSION

It makes sure that the user clicks on the options before he picks a preset.
This preset now just won't do anything if user don't check the options, so there we go.

limonspb:
looks good, please squash and ready to go

limonspb (approved):

limonspb (approved):

JyeSmith:
Sorry for the spam... I think Im done 😄

sugaarK:
Amazing @JyeSmith

sugaarK:
didn't think about this at the time but it should be a new preset in the new 4.4 folder and leave the 4.3 version as it was

limonspb:
> didn't think about this at the time but it should be a new preset in the new 4.4 folder and leave the 4.3 version as it was

Could be ok here, he mentioned the BF version in the options. Options still needed here cuz even in 4.4 not everyone needs MSP maybe?

JyeSmith:
Would that then mean there are 2 files to maintain?  It would be great if it could stay and there was only 1.

PR #297, opened 2022-10-12
add 4.4 preset for HDZERO vtxs including MSP VTX setup 

sugaarK:
added a 4.4 preset for the range of HDZERO vtxs that includes the ability to setup up the MSP VTX serial settings from the preset...  

more information about the MSP VTX settings and use in this PR https://github.com/betaflight/betaflight/pull/11705

### DIFF
--- a/presets/4.4/vtx/HDZero.txt
+++ b/presets/4.4/vtx/HDZero.txt
@@ -1,0 +1,123 @@
+#$ TITLE: HDZero VTXs
+#$ FIRMWARE_VERSION: 4.4
+#$ CATEGORY: VTX
+#$ STATUS: COMMUNITY
+#$ KEYWORDS:  vtx, vtx table, HDzero, divimath, SA 2.1, digital, whoop, whoop lite, race v1, race v2, freestyle, tx5m.1, MSP VTX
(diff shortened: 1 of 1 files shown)


KarateBrot (approved):

haslinghuis:
22 seconds :-)

limonspb:
> In the Options is the Freestyle V2 Considered "All VTXs (except unlocked freestyle and TX5m.1)" Or is the V2 treated just like a V1 freestyle here?
> 
> Just looking for clarification on what to select for a unlocked freestyle v2.

@sugaarK 

also padding for options! :)

sugaarK:
> In the Options is the Freestyle V2 Considered "All VTXs (except unlocked freestyle and TX5m.1)" Or is the V2 treated just like a V1 freestyle here?
> 
> Just looking for clarification on what to select for a unlocked freestyle v2.

no idea I don't Own one.. if you post the vtx table and we can see

thespova:
VTX Table is the same for unlocked and locked v1 Freestyle.  I'm going to assume the unlocked config is just for the power output options and that it applies to v1 and v2 freestyle when unlocked.

PR #300, opened 2022-10-16
Update defaults.txt for Betaflight 4.4

TehllamaFPV:
New Antigravity Settings:
# -- Antigravity (default) --
set anti_gravity_gain = 80
set anti_gravity_cutoff_hz = 5
set anti_gravity_p_gain = 100

This is part of the dependency chain to enable new Tune and TUNE_OTHER presets to exist for Betaflight 4.4

Revised Antigravity settings require these changes - placed the defaults (as of 2022.10.14 build), 
otherwise the OFFICIAL default tune remains unchaged,.

PR #301, opened 2022-10-16
Create Tehllama Llama 5in Race (Betaflight 4.4) Preset [Unified Multivoltage 3/4/5/6-S Tune for 1600-2750KV]

TehllamaFPV:
Preset for Betaflight 4.4 (Tested using the 2022.10.14 Nightly)

Once user sets Motor KV, Filter Flavor (non-RPM, Safe, Spicy), Feature Selection...
 this will automatically set up an array of 3S, 4S, 5S, and 6S tunes based on motor KV with matching filters.
 Master Slider increments are locked to 0.05 increments, and end user adjustable after flashing preset

// Key Update Milestones
 - Will update on feature freeze date
 - Will update on all namespace changes related to Tune profiles or Filtering
 - Will update on Release Candidate 0
 - Will review on all release candidates

// Dependency Chains Identified [This is why the error checker will throw errors until these items are present in MASTER
/// See Pull Request #300 
 - * presets/4.4/tune/defaults.txt 
 !!! This file requires an update due to Antigravity Changes !!!
Existing file - can directly port into the 4.4 folder, and replace the Antigravity block with the following:
-- Antigravity (default) --
set anti_gravity_gain = 80
set anti_gravity_cutoff_hz = 5
set anti_gravity_p_gain = 100

#-- && Filter Defaults
/// See Pull Request #302 
 - Relies on an update to the following files:
 - * presets/4.4/filters/defaults.txt and presets/4.4/filters/basic_no_rpm_*****.txt
  (Existing - can directly port the 4.3 Default Filters - just add ', 4.4' to the VERSION text so node.js can parse it
// Can create additional pull requests to support these

This is a progression of the existing trio of Llama 5in Race Presets - Merged into a single preset, with Spicy/NonSpicy/nonRPM versions enabled for Filter and Tune

The really long list of Motor KV options provides Master Slider scaling for all four voltages, applies motor_output limiting to emulate Freedom_Spec with 3S packs at 2000KV and above, and also applies motor output limiting on 6S and 5S high KV setups which would exceed 48,000RPM at realistic voltages (transonic tip speeds that damage batteries and do not add thrust).

Optional Features section includes MaxCellVoltage options (For Freedom Spec, allows for 4.35V/cell charging without OSD errors)
Disables VBat_Sag_Compensation for FreedomSpec by default (this can be removed to keep this as a valid 'tune' preset, but this is an edge case.
Dynamic Idle is enabled by default, but still requires bidirectional DShot, as do the RPM Filter Enabled (Safe/Spicy) versions

### DIFF
--- a/presets/4.4/other/Tehllama_5in_Race_Multivoltage.txt
+++ b/presets/4.4/other/Tehllama_5in_Race_Multivoltage.txt
@@ -1,0 +1,694 @@
+#$ TITLE: Tehllama 5" Race Multivoltage
+#$ FIRMWARE_VERSION: 4.4
+#$ CATEGORY: OTHER
+#$ STATUS: EXPERIMENTAL
+#$ KEYWORDS: 533, Freedom, FreedomSpec, Tune, Spec, Race, Llama, MultiVoltage, Texas, BMS, 3S, 5in, AutoSelect, MM, Omni
+#$ AUTHOR: Daniel Appel / Tehllama
+#$ DESCRIPTION: This is a multi-voltage tune built for 5in Race Builds in the 400-575g AUW range. 
+#$ DESCRIPTION: Use the Options to select KV - 1600-2750KV are supported for the entire 3S-6S voltage range
+#$ DESCRIPTION: ALL FOUR Tune Profiles are used, with tunes Auto-Selected for 6S, 5S, 4S, and 3S (respectively) 
+#$ DESCRIPTION:
+#$ DESCRIPTION: Select the Motor KV Range from the Drop-Down Menu that most closely matches your craft
+#$ DESCRIPTION: Default is for ~1950KV motors, for example Freedom Spec 2207 1960KV 533/XNOVA Motors
+#$ DESCRIPTION: Special Thanks to Stephen 'Supafly' Wright for Master Slider Calculator assistance
+#$ DESCRIPTION: 
+#$ DESCRIPTION: This tune works best with bidirectional DShot enabled, running RPM and Dynamic Idle Features
+#$ DESCRIPTION: This tune expects ESCs configured to 48kHz PWM Frequency, 23°-27°/MedHigh timing and 0.38 Rampup
+#$ DESCRIPTION: Using more aggressive ESC settings is NOT advised if using higher cell count batteries
+#$ DESCRIPTION: 
+#$ DESCRIPTION: This tune will auto-select profiles based on battery voltage at plugin.
+#$ DESCRIPTION: When 3S packs are plugged in, this will auto-select an aggressive spec racing tune on Profile 3
+#$ DESCRIPTION: At higher motor KV, MotorOutput Limiting is used to scale these to Freedom Spec RPM 
+#$ DESCRIPTION: 3S batteries on craft with <1950KV motors experience very reduced performance
+#$ DESCRIPTION: 
+#$ DESCRIPTION: 6S, 5S, and 4S will run conservative open race tunes 
+#$ DESCRIPTION: Small motor output limits are applied for EXTREME Battery-KV combinations
+#$ DESCRIPTION: Combinations that spin 5" propellers in transonic regimes (i.e. 6S >2200KV and 5S >2650KV)
+#$ DESCRIPTION: 
+#$ DESCRIPTION: !!! Strongly recommend a full chip erase reflash if alternative tunes are desired after installing this preset!!!
+#$ DESCRIPTION: Extensive testing has been done across 83+ builds, however not every craft will run best on this tune.
+#$ WARNING: Use at your own risk.  Verify Correct ESC Settings are used.  
+#$ DISCUSSION: https://github.com/betaflight/firmware-presets/pull/301
+
+## Filters - Choose RPM Enabled (Spicy or Non-Spicy) or non-RPM (Super-Safe)
+## Filters - by Profile x4 (Enables 6S, 5S, 4S, and 3S Spec Tune)
+## Tune - TPA, iTerm, dMin, boost, & AutoCellSelect Parameters
+## Tune - PI, D, I, Dmax, pitch PI/D, FF Offsets
+## KV Selection - Set MM Slider for ALL profiles
+## KV Selection - Set Motor Output Limits where applicable by profile
+## Feature Selection - VBatSagComp, MaxCellVoltage
+
+#$ FORCE_OPTIONS_REVIEW: TRUE
+
+#$ OPTION_GROUP BEGIN: Filters
+    #$ OPTION BEGIN (CHECKED): Apply RPM Enabled SAFE Filters to ALL profiles (Requires Bidirectional DShot)
+		profile 0
+			#$ INCLUDE: presets/4.3/filters/defaults.txt
+			set dterm_lpf1_dyn_min_hz = 101
+			set dterm_lpf1_dyn_max_hz = 266
+			set dterm_lpf1_dyn_expo = 10
+			set dterm_lpf1_static_hz = 0
+			set dterm_lpf2_static_hz = 202
+			set yaw_lowpass_hz = 111
+			set simplified_dterm_filter_multiplier = 135
+			set simplified_dterm_filter = OFF
+			simplified_tuning apply
+		profile 1
+			#$ INCLUDE: presets/4.3/filters/defaults.txt
+			set dterm_lpf1_dyn_min_hz = 101
+			set dterm_lpf1_dyn_max_hz = 255
+			set dterm_lpf1_dyn_expo = 9
+			set dterm_lpf1_static_hz = 0
+			set dterm_lpf2_static_hz = 202
+			set yaw_lowpass_hz = 111
+			set simplified_dterm_filter_multiplier = 135
+			set simplified_dterm_filter = OFF
+			simplified_tuning apply
+		profile 2
+			#$ INCLUDE: presets/4.3/filters/defaults.txt
+			set dterm_lpf1_dyn_min_hz = 101
+			set dterm_lpf1_dyn_max_hz = 244
+			set dterm_lpf1_dyn_expo = 8
+			set dterm_lpf1_static_hz = 0
+			set dterm_lpf2_static_hz = 202
+			set yaw_lowpass_hz = 111
+			set simplified_dterm_filter_multiplier = 135
+			set simplified_dterm_filter = OFF
+			simplified_tuning apply
+		profile 3
+			#$ INCLUDE: presets/4.3/filters/defaults.txt
+			set dterm_lpf1_dyn_min_hz = 101
+			set dterm_lpf1_dyn_max_hz = 233
+			set dterm_lpf1_dyn_expo = 5
+			set dterm_lpf1_static_hz = 202
+			set dterm_lpf2_static_hz = 
+			set yaw_lowpass_hz = 111
+			set simplified_dterm_filter_multiplier = 135
+			set simplified_dterm_filter = OFF
+			simplified_tuning apply
+		# Filter Master
+			set dyn_notch_count = 3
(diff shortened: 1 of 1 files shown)


TehllamaFPV:
Changed type to 'OTHER' - since simply replacing the tune preset in whichever profile the user has active would result in the other three profiles still being present, and wanting to make themselves active when the battery cell count they are intended to trigger on is plugged in.

presets/4.4/other/Tehllama_5in_Race_Tune_1600-2750KV_Freedom_MultiVoltage_3S-4S-5S-6S.txt
```diff
@@ -0,0 +1,702 @@
...
+## KV Selection - Set MM Slider for ALL profiles
+## KV Selection - Set Motor Output Limits where applicable by profile
+## Feature Selection - VBatSagComp, MaxCellVoltage
+
+#$ FORCE_OPTIONS_REVIEW: TRUE
+
+#$ OPTION_GROUP BEGIN: Filters
+#$ OPTION BEGIN (CHECKED): Apply RPM Enabled SAFE Filters to ALL profiles (Requires Bidirectional DShot)
+profile 0
+#$ INCLUDE: presets/4.3/filters/defaults.txt
```
haslinghuis, line 47: Should we just make a presets/4.4/filters/defaults.txt (changed or not)?

TehllamaFPV: Discussion in the other thread - @limonspb is correct that there are no changes, so referencing the 4.3 file location doesn't have any ill effects on this

haslinghuis (approved):

limonspb:
sorry @haslinghuis we have a conversation in Slack about this PR, plz don't merge yet.

TehllamaFPV:
> sorry @haslinghuis we have a conversation in Slack about this PR, plz don't merge yet.

Concur - I've realized there are a few odd behaviors I need to track down (antigravity not behaving consistently, and change the name so it's not so brutally truncated on phone screens... at a minimum)

limonspb:
> > sorry @haslinghuis we have a conversation in Slack about this PR, plz don't merge yet.
> 
> Concur - I've realized there are a few odd behaviors I need to track down (antigravity not behaving consistently, and change the name so it's not so brutally truncated on phone screens... at a minimum)

don't forget indentation :)

haslinghuis:
Was not part of the conversation. But why we need indentation. Are we going back to COBOL times? :grinning: Sure it's more readable in review but this is machine code.

limonspb:
> Was not part of the conversation. But why we need indentation. Are we going back to COBOL times? 😀 Sure it's more readable in review but this is machine code.

Yeah, easier to read the code for options and option groups open/close tags

haslinghuis:
Yeah maybe we could migrate to yaml syntax

limonspb:
> Yeah maybe we could migrate to yaml syntax in distance future.

Nah, the current syntax is cool cuz every preset is also a correct CLI list of commands. Can copy paste pieces back and forth to/from your diff/dump.

haslinghuis:
COBOL was about having code at specific places. https://www.tonymarston.net/cobol/cobolstandards.html

If we just like to have configuration like C defines there is really no need to use indentation other then being easier to read. But then we should use indentation on other directives too.

limonspb:
> COBOL was about having code at specific places. https://www.tonymarston.net/cobol/cobolstandards.html
> 
> If we just like to have configuration like C defines there is really no need to use indentation other then being easier to read. But then we should use indentation on other directives too.

Like in many other languages the indentation is just a coding style for easy read. And is used for code blocks and nested code blocks. No difference with presets code blocks for options and option groups.

haslinghuis:
Totally agree,

Having all cli commands indented will make it much more readable, not only for option directive.

limonspb:
> Totally agree,
> 
> For consistency having all cli commands indented will make it much more readable, not only for option directive.

It's nice to visually divide commands outside of any blocks from commands inside the blocks.

```
Set A
Set B
Set C

#$ OPTION BEGIN
    Set D
    Set E
#$OPTION END

Set F
Set G
```

haslinghuis:
Sure only commands that are within a directive should be indented but have nothing against all commands being indented (like markup examples in documents)

limonspb:
> Sure only commands that are within a directive should be indented but have nothing against all commands being indented (like markup examples in documents)

I don't quite understand what you talking about, please provide example

haslinghuis:
Just like it this way. Directives on column 1, commands on column 5 :-) 

```
    Set A
    Set B
    Set C

#$ OPTION BEGIN
    Set D
    Set E
#$OPTION END

    Set F
    Set G
```

limonspb:
> Just like it this way. Directives on column 1, commands on column 5 :-)
> 
> ```
>     Set A
>     Set B
>     Set C
> 
> #$ OPTION BEGIN
>     Set D
>     Set E
> #$OPTION END
> 
>     Set F
>     Set G
> ```

This way CLI commands outside of option seems to be on the same level as commands inside options. Will be harder to read.

haslinghuis:
That's just our opinion and different interpretation.

As I missed the conversation: Are there directives inside other directives, As they should be indented too,

limonspb:
> That's just our opinion and different interpretation.
> 
> As I missed the conversation: Are there directives inside other directives, As they should be indented too,

It's a common thing in any language that you add the next identation level only when you open a new block.

We don't do:

```
    let A;
If B {
    C;
}
```

Instead we do:
```
let A;
If B {
    C;
}
```

haslinghuis:
Agree, as long it's not part of a directive if should be column 1. Commands within a directive should use column 5. If there are sub directives within a directive it should move to column 9 instead of 5 likewise with sub sub directives (there are none so far afaik.

haslinghuis (approved):

TehllamaFPV:
I wound up shoving things over to Col13 just so I could have some visual clarity involved with multiple profiles, mostly because those nested within option groups and option sets, any indentation is helpful.
For tunes that sane people actually implement, I think @haslinghuis is spot on about using 1/5/9 as recommended

It's no longer a 700 line preset, the indentation alone let me clear out a bunch of hanging carriage returns, so this is the way forward, and 1000% my vote (even though my keypress count for the day is through the roof)

limonspb (changes requested):
Title and file name are still long.
Some option names are also way too long, wondering how they look on the phone.
simplified tuning apply can be mentioned once just at the very end of a preset?
Identation should be 4 spaces, not a tab symbol, its a setting in any text editor.

TehllamaFPV:
I've found that the sliders behave very inconsistently if placed, but not applied before switching profiles - the gyro filter slider will apply, but the rest tend to be practically there in real PID values, but don't actually appear as slider values, just manually entered ones in configurator... when explicitly called before changing profiles, it seems to work just fine.

I'm on a different computer, I might have retained those notepad++ presets...
It still runs pretty elegantly (local repo node.js builds and runs, preset appears to be working with tab)

I'll break out my phone and start playing with filename options - honestly this is a sufficiently niche thing that I can live with a bit of poor readability, I'm probably walking people through this process in real use.

limonspb:
> I've found that the sliders behave very inconsistently if placed, but not applied before switching profiles - the gyro filter slider will apply, but the rest tend to be practically there in real PID values, but don't actually appear as slider values, just manually entered ones in configurator... when explicitly called before changing profiles, it seems to work just fine.
> 
> I'm on a different computer, I might have retained those notepad++ presets... It still runs pretty elegantly (local repo node.js builds and runs, preset appears to be working with tab)
> 
> I'll break out my phone and start playing with filename options - honestly this is a sufficiently niche thing that I can live with a bit of poor readability, I'm probably walking people through this process in real use.

ok, got it about simplified tuning apply.
Well if we add a preset in the official repo, we should try our best to improve UI and user experience with this preset, without the necessity of interacting with the author :)
if your preset is too complicated for most users, we should break it in multiple presets, make less options, for example:
combine tune and filters options.
Separate RPM and NO-RPM in different presets or even jsut remove non-RPM
Include dynamic idle without option. Dynamic idle is good, why anyone turns it off
and other things.

TehllamaFPV:
The idea is to make it a very set-it-and-forget-it type tune... The Option and OptionGroups are awesome.

Definitely agree - having this as an 'OTHER' preset frees that up a fair bit too - I can set stuff ON by default, for most users it might as well be hard coded in that case.
I don't see this as being a logical 'OFFICIAL' preset, because it does a lot of complicated things, but I feel like community makes sense because this is the most idiot-proof tune I'm aware of that can be ran on a race quad... and I'm usually the universe's better idiot when it comes to testing that.
There are exactly two quads (not mine) I have in mind when including all of those levers... I have tested the 4.3 version of this tune on multi-shot only ESCs, and it still parties

I've got this on 7 birds already,
New builds, it's as easy as grabbing this, rate preset, RC_Link, and VTX presets, then just making sure the motors/config tabs look right on the new builds.
The old ones, I can just smash this and the ratebank presets and be completely done after reflashing and adding the stale diff file.

Still GAME CHANGING to have presets for all this; I can put this hilariously complicated multi-option thing on, and apply it correctly in less time than it takes to just reflash the ESCs; I can't get over that

limonspb:
> The idea is to make it a very set-it-and-forget-it type tune... The Option and OptionGroups are awesome.
> 
> Definitely agree - having this as an 'OTHER' preset frees that up a fair bit too - I can set stuff ON by default, for most users it might as well be hard coded in that case. I don't see this as being a logical 'OFFICIAL' preset, because it does a lot of complicated things, but I feel like community makes sense because this is the most idiot-proof tune I'm aware of that can be ran on a race quad... and I'm usually the universe's better idiot when it comes to testing that. There are exactly two quads (not mine) I have in mind when including all of those levers to run this RPM-less... I have tested the 4.3 version of this tune on multi-shot only ESCs, and it still parties.
> 
> I've got this on 7 birds already, New builds, it's as easy as grabbing this, rate preset, RC_Link, and VTX presets, then just making sure the motors/config tabs look right on the new builds. The old ones, I can just smash this and the ratebank presets and be completely done after reflashing and adding the stale diff file.
> 
> Still GAME CHANGING to have presets for all this; I can put this hilariously complicated multi-option thing on, and apply it correctly in less time than it takes to just reflash the ESCs; I can't get over that

A new preset becomes community only after hanging as experimental for some time in the official repo. Exceptions can be made for small clear presets or for minor updates of the preset files.

The long title - just nobody going to see it in most cases cuz it's too long. Long options names will just look not nice.

presets/4.4/other/Tehllama_5in_Race_MultiTune_1600-2750KV_3-6S.txt
```diff
@@ -0,0 +1,693 @@
+#$ TITLE: Tehllama 5in Race MultiVoltage Tune - 1600-2750KV 4S/5S/6S (Open) & 3S (Spec)
```
sugaarK, line 1: way to long

TehllamaFPV:
Changed to EXPERIMENTAL - corrected indentation in two locations.
The only part of the option names that gets clipped on the phones and low resolution tablets I've tried are the ' (Bidir DShot required )' elements, which I am OK with.
I still want to push to have this reach the Community status by RC1, but I won't be unhappy if it remains experimental all the way to the release

presets/4.4/other/Tehllama_5in_Race_Multi.txt
```diff
@@ -0,0 +1,694 @@
+#$ TITLE: Tehllama 5" Race Multi
```
limonspb, line 1: Multivoltage :)

limonspb (approved):

PR #302, opened 2022-10-18
BF4.4 Filter Presets :: Porting defaults, basic, and karate 4.3 tune presets to 4.4

TehllamaFPV:
** Breaking ALL the rules **
// Primarily for dependency chain reasons

Did a simple copy over, then Find & Replace All 4.3 to 4.4 for each file
Made minor spelling corrections in Karate Preset

BMI270 presets NOT ported over - would need clarification on OSR4 mode, default behaviors, and what most closely works with the wider range of presets to emulate MPU6000-like performance with the BMI270 hardware.

Dependency chain operation works with this simple port-over and FIRMWARE_VERSION iteration++, appears to work fine.

### DIFF
--- a/presets/4.4/filters/basic_no_rpm_clean.txt
+++ b/presets/4.4/filters/basic_no_rpm_clean.txt
@@ -1,0 +1,46 @@
+#$ TITLE: 4.4 Filter settings for CLEAN build WITH NO RPM FILTERING.
+#$ FIRMWARE_VERSION: 4.4
+#$ CATEGORY: FILTERS
+#$ STATUS: OFFICIAL
+#$ KEYWORDS: filter, filters, filtering, clean, no rpm, basic, betaflight
+#$ AUTHOR: Betaflight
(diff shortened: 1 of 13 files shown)


haslinghuis (approved):

limonspb (changes requested):
filters are the same between 4.3 and 4.4 probably?
In this case no need for separate files

TehllamaFPV:
I can confirm, the filters are the same appearance-wise in the configurator.
I realize I introduced a syntax error when I thought node.js needed to have them in the same local file structure...

We can close this PR unless some fruitful discussion about how to incorporate this best makes sense - my personal vote would be the answer below, but duplicating files may not be the worst answer when it comes to eventual version control behaviours.

The alternative would be appending the #$ FIRMWARE_VERSION: 4.4 to the existing 4.3 files so that that Preset system will search those (instead of those appearing absent for any users running presets alongside development 4.4 builds in the interim

limonspb:
4.3->4.4 filters are the same, so can add 4.4 firmware tag to 4.3 filter presets

PR #304, opened 2022-10-23
Added FRSKY FPort presetting

HThuren:
Inspired by frsky_sbus preset, but with singlewire FPort:
set serialrx_provider = FPORT
set serialrx_inverted = ON
set serialrx_halfduplex = ON


limonspb:
Hi
As far as I can see this preset is pretty much identical to frsky one besides a few CLI commands at the beginning.
Maybe it makes sense to just modify the existing frsky preset and extend it with the FPORT option?
Like add FPORT to title and keywords and add an option.

HThuren:
Fine suggestion, I'll do that and close this PR

HThuren:
Now at PR #305

PR #312, opened 2022-11-02
Karate tunes 4.3->4.4 port

limonspb:
This is just a simple port from 4.3 to 4.4 for the Karate tunes/filters by @sugaarK. Mostly copy-paste.
Filters stay the same so far.
Tunes require new AG settings so leaving them at defaults until further testing. Marking 4.4 tunes as experimental.

Read more about the new 4.4 AG:
https://github.com/betaflight/betaflight/pull/11679

Leaving discussion links the same as for 4.3


sugaarK (approved):
Stock anti grav works good with all These tunes so no need to change

PR #334, opened 2022-12-15
update all rc presets to 4.4 tag

sugaarK:
added 4.4 tag to all rc link presets and fixed default and moved to 4.4 folder


limonspb (approved):
Ha, good catch, i was 100% sure i did that for all presets, but missed RC_LINK somehow 0_o

PR #335, opened 2022-12-16
Tiny_karate: added motor_poles = 12

limonspb:
The more i think about motor_poles the more i lean towards that we should allow it outside of the options in TUNE/filters presets.
But only in the cases like tinywhoops, for example or Tiny Trainers. When the chances of swapping motors to a different pole count is pretty much zero.


sugaarK:
I added this on my whoop time as an option but checked

limonspb:
> I added this on my whoop time as an option but checked

yeah i saw that and decided to let it go even tho its agains't the current rules lol

haslinghuis (approved):

sugaarK (approved):

sugaarK:
> The more i think about motor_poles the more i lean towards that we should allow it outside of the options in TUNE/filters presets.
> But only in the cases like tinywhoops, for example or Tiny Trainers. When the chances of swapping motors to a different pole count is pretty much zero.

This is my take.. make it a checked option

PR #343, opened 2023-02-04
4.3 to 4.4 UAV Tech Presets Update

spatzengr:
Updated my 7 presets coming from 4.3 to 4.4.  PID Settings are the same.  Mostly changed the header information to be more user-friendly.  I did add filter options.  Once merged I have 4 new presets to add for which I will create a separate PR for each.


index.json
```diff
@@ -134,8 +133,10 @@
   },
   "uniqueValues": {
     "firmware_version": [
+      "3.5",
```
haslinghuis, line 136: Configurator does support >= 4.0 - this would need configurator 10.8 not compatible with 4.4 firmware.

index.json
```diff
@@ -177,17 +182,25 @@
...
+      "2.5W",
+      "2022",
       "20ms",
+      "2500mW",
       "250Hz",
+      "2in",
       "2ms",
-      "3 inch",
+      "2S",
+      "3 Inch",
```
haslinghuis, line 199: Please try to use the same name - either `in` or` inch` - and same capitalization as this is confusing.

index.json
```diff
@@ -699,18 +689,57 @@
...
-        "default",
         "betaflight"
       ],
       "author": "Betaflight",
       "priority": 100
     },
+    {
+      "fullPath": "presets/4.3/filters/ctzsnooze_race.txt",
+      "hash": "3cd6b38b21527d3acb44b6e73e0df0816f9bef36ad1c4a83644d51a903aa579b",
+      "title": "ctzsnooze's race filter settings",
```
haslinghuis, line 700: Maybe we should use Title Case in all titles.

Just trying to make the content more consistent.

spatzengr:
index update not needed and causing confusion.

PR #353, opened 2023-02-06
UAV_tech_Cinewhoop

spatzengr:
Updated cinewhoop preset.


limonspb:
looks good, similar question, if you wanna setup poles count since we started to allow this setting for filters if they set RPM filtering on.

limonspb (approved):

PR #354, opened 2023-02-06
UAV_tech_Micro

spatzengr:
Updated toothpic/micro preset.


presets/4.4/tune/uav_tech/UAV_tech_Micro.txt
```diff
@@ -0,0 +1,132 @@
...
+#$ DESCRIPTION: - [Preset Overview Video](https://youtu.be/dEuNa-V6pys)
+#$ DESCRIPTION:
+#$ DESCRIPTION:
+#$ DESCRIPTION: Preset for this class of quadcopters:
+#$ DESCRIPTION: -----------
+#$ DESCRIPTION: <img src="https://theuavtech.com/wp-content/uploads/2020/10/Thoothpic-Header-3.jpg" width="350px"/>
+#$ DESCRIPTION:
+#$ DESCRIPTION: Description:
+#$ DESCRIPTION: -----------
+#$ DESCRIPTION: Base tune for a micro quadcopters; anything from 2" ("Toothpic") to 4".  This tune is good for 2S or 6S batteries.
```
haslinghuis, line 21: 2S or 6S batteries :)

presets/4.4/tune/uav_tech/UAV_tech_Micro.txt
```diff
@@ -0,0 +1,129 @@
+#$ TITLE: UAV Tech - Micro (2" to 4")
+#$ FIRMWARE_VERSION: 4.4
+#$ CATEGORY: TUNE
+#$ STATUS: EXPERIMENTAL
+#$ KEYWORDS: 2S, 3S, 4S, 6S, toothpic, 2in, 2.5in, 3in, 3.5in, 4in
```
limonspb, line 5: 6S?

limonspb (approved):
looks good.
Do you want to also set the poles count?
or all toothpicks got different motors with different poles count?

I figured we should allow poles count setting for filters/pids presets.

spatzengr:
> looks good. Do you want to also set the pole count? or all toothpicks got different motors with different pole count?

I don't think I want to touch Pole Count for Micros.  It varies.  I have run both ways and it is not the end of the world if it is to be 12 and it is set at 14.  It just means the RPM notch isn't as effective at attenuating noise, but I'm leaving 2 of the DN filters enabled, so any peaks that are missed the DN will pick them up anyways.

PR #356, opened 2023-02-10
3 4 Inch Tune for 4.3 Presets

SupaflyFPV:
Here is a new preset for 3 to 4 Inch quads for 4.3, with included PIDs, Filters, Radio and Rates setup for freestyle

### DIFF
--- a/presets/4.3/tune/SupaflyFPV_Freestyle_3_4_Inch_EasyTune.txt
+++ b/presets/4.3/tune/SupaflyFPV_Freestyle_3_4_Inch_EasyTune.txt
@@ -1,0 +1,262 @@
+#$ TITLE: SupaflyFPV Freestyle 3/4 Inch
+#$ FIRMWARE_VERSION: 4.3
+#$ CATEGORY: TUNE
+#$ STATUS: EXPERIMENTAL
+#$ KEYWORDS: Supafly, SupaflyFPV, 4S, 6S, Freestyle, 3 inch, 4 inch, 3", 4"
+#$ AUTHOR: SupaflyFPV
+
+#$ PARSER: MARKED
+
+#$ DESCRIPTION: <a href="https://supaflyfpv.com"><img src="https://supaflyfpv.com/wp-content/uploads/2022/06/Logo-Cropped192.svg" width="220px" style="margin-left: auto; margin-right: auto; display: block;"/></a>
+#$ DESCRIPTION:
+#$ DESCRIPTION: <br>
+#$ DESCRIPTION:
+#$ DESCRIPTION: ### SupaflyFPV 3/4 Inch Freestyle Preset
+#$ DESCRIPTION:
+#$ DESCRIPTION: <br>
+#$ DESCRIPTION:
+#$ DESCRIPTION: Choose from the options in drop-down menu above:
+#$ DESCRIPTION:
+#$ DESCRIPTION: <br>
+#$ DESCRIPTION:
+#$ DESCRIPTION: * For optimisation and tips [use this custom PID calculator](https://supaflyfpv.com/betaflight-setup/)
+#$ DESCRIPTION: * Check what gyro your flight controller is using
+#$ DESCRIPTION: * RPM Filtering recommended on compatible ESC or de-select if not available
+#$ DESCRIPTION: * Pick your Radio Link or search in presets if not found
+#$ DESCRIPTION: * Try some freestyle rates modeled on the 3 main styles...
+#$ DESCRIPTION: 
+#$ DESCRIPTION: <br>
+#$ DESCRIPTION:
+#$ DESCRIPTION: Enjoy!
+#$ DESCRIPTION:
+#$ DESCRIPTION: <br>
+#$ DESCRIPTION:
+#$ DESCRIPTION: Support, feedback and chat:
+#$ DESCRIPTION:
+#$ DESCRIPTION: <br>
+#$ DESCRIPTION:
+#$ DESCRIPTION:*[FACEBOOK SUPPORT/FEEDBACK](https://www.facebook.com/groups/supaflyfpv/)
+#$ DESCRIPTION:
+#$ DESCRIPTION: <br>
+#$ DESCRIPTION:
+#$ DESCRIPTION:*[DISCORD SUPPORT/FEEDBACK](https://discord.gg/Gs5nTx4rE2)
+#$ DESCRIPTION:
+#$ DESCRIPTION: <br>
+#$ DESCRIPTION: 
+
+#$ INCLUDE: presets/4.3/tune/defaults.txt
+
+# -- Antigravity --
+
+set anti_gravity_gain = 5000
+
+#$ OPTION_GROUP BEGIN:PIDs
+    #$ OPTION BEGIN (CHECKED): With HD Camera
+
+        set simplified_master_multiplier = 100
+        set simplified_i_gain = 75
+        set simplified_d_gain = 110
+        set simplified_pi_gain = 110
+        set simplified_dmax_gain = 0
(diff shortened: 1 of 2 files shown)


limonspb (changes requested):
a few notes.
1. Fresh presets should be experimental.
2.  Option groups do not work like that. The only purpose for them is to be a container for options.
For example, you have an option group for rates. And inside the option group, you have included the rate defaults. It means that default rates will be applied for everyone no matter what. So including rates default should be inside option, not just option group.
3. You have the same for filters.
4. Also, there is a slight mess with indentations. This is how it should be, 1 level - 4 spaces:
```
#$ OPTION_GROUP BEGIN: this group name
    #$ OPTION BEGIN (UNCHECKED): <Option1 name>
        CLI payload strings
    #$ OPTION END
    #$ OPTION BEGIN (UNCHECKED): <Option2 name>
        CLI payload strings
    #$ OPTION END
#$ OPTION_GROUP END
```
5. I see that filters do use sliders here and there, but then don't have "simplified_tuning apply" after. The safet way would be to use `simplified_tuning apply` at the very end of the preset only to make sure it covers everything.
6. you have `set anti_gravity_gain = 100` in the filters option.
7. Checker crashes, because there is no "presets/4.4/rc_link/defaults.txt", there is just "presets/4.3/rc_link/defaults.txt" which works for both 4.3 and 4.4

SupaflyFPV:
> a few notes.
> 
> 1. Fresh presets should be experimental.
> 2. Option groups do not work like that. The only purpose for them is to be a container for options.
>    For example, you have an option group for rates. And inside the option group, you have included the rate defaults. It means that default rates will be applied for everyone no matter what. So including rates default should be inside option, not just option group.
> 3. You have the same for filters.
> 4. Also, there is a slight mess with indentations. This is how it should be, 1 level - 4 spaces:
> 
> ```
> #$ OPTION_GROUP BEGIN: this group name
>     #$ OPTION BEGIN (UNCHECKED): <Option1 name>
>         CLI payload strings
>     #$ OPTION END
>     #$ OPTION BEGIN (UNCHECKED): <Option2 name>
>         CLI payload strings
>     #$ OPTION END
> #$ OPTION_GROUP END
> ```
> 
> 5. I see that filters do use sliders here and there, but then don't have "simplified_tuning apply" after. The safet way would be to use `simplified_tuning apply` at the very end of the preset only to make sure it covers everything.
> 6. you have `set anti_gravity_gain = 100` in the filters option.
> 7. Checker crashes, because there is no "presets/4.4/rc_link/defaults.txt", there is just "presets/4.3/rc_link/defaults.txt" which works for both 4.3 and 4.4

OK I'll see to those changes...I actually only intended to PR the 4.3 version of this preset as we discussed - one at a time...not sure how I managed to also upload the 4.4 version...shall I remove that one or modify it as well?

limonspb:
> > a few notes.
> > 
> > 1. Fresh presets should be experimental.
> > 2. Option groups do not work like that. The only purpose for them is to be a container for options.
> >    For example, you have an option group for rates. And inside the option group, you have included the rate defaults. It means that default rates will be applied for everyone no matter what. So including rates default should be inside option, not just option group.
> > 3. You have the same for filters.
> > 4. Also, there is a slight mess with indentations. This is how it should be, 1 level - 4 spaces:
> > 
> > ```
> > #$ OPTION_GROUP BEGIN: this group name
> >     #$ OPTION BEGIN (UNCHECKED): <Option1 name>
> >         CLI payload strings
> >     #$ OPTION END
> >     #$ OPTION BEGIN (UNCHECKED): <Option2 name>
> >         CLI payload strings
> >     #$ OPTION END
> > #$ OPTION_GROUP END
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > 5. I see that filters do use sliders here and there, but then don't have "simplified_tuning apply" after. The safet way would be to use `simplified_tuning apply` at the very end of the preset only to make sure it covers everything.
> > 6. you have `set anti_gravity_gain = 100` in the filters option.
> > 7. Checker crashes, because there is no "presets/4.4/rc_link/defaults.txt", there is just "presets/4.3/rc_link/defaults.txt" which works for both 4.3 and 4.4
> 
> OK I'll see to those changes...I actually only intended to PR the 4.3 version of this preset as we discussed - one at a time...not sure how I managed to also upload the 4.4 version...shall I remove that one or modify it as well?

not a big deal if they are both just copy-paste except anti-gravity.

SupaflyFPV:
Ok hopefully that has corrected most of those...

I cleared up some of the 'defaults' positioning mentioned, however some of the filters is intentionally outside a group option hence the filters defaults triggering automatically.

Got it re the spacing.

SupaflyFPV:
...I only recently started using the sliders because everyone has their own idea about gyro low pass filters so let them do what they want...

But my question is: for CLI command with sliders on - do I need to add in the filter frequencies 'and' the filter slider value? Can I just use one or the other?

limonspb:
> ...I only recently started using the sliders for this community tune because everyone has their own idea about gyro low pass filters so let them do what they want...
> 
> But my question is: for CLI command with sliders on - do I need to add in the filter frequencies 'and' the filter slider value? Can I just use one or the other?

So if you want to use filters sliders (sliders are ON by default)
1. set slider value
2. set some LPF to zero if you want some filters to be off
3. use simplified_tuning apply

if you want to use your custom values for filters:
1. set simplified_gyro_filter = OFF
2. set your custom values for filters that are different from defaults

keep in mind that gyro filter sliders and dterm filter sliders are separate, and can be on/off independantly.


PS:
- i still see you got identations wrong in lots of places.
- i would keep your "defaults" filter settings outside of the option group, because option group is to just group options.

limonspb:
PS: presets are experimental also because it's basically a new src code that can be messed up.

SupaflyFPV:
ok done. I removed the base filter setup from the Filters group, let me know if thats what you meant...

Do you want commits squashed for review or leave them seperated?

limonspb:
> ok done. I removed the base filter setup from the Filters option group, let me know if thats what you meant...
> 
> Do you want commits squashed for review or leave them seperated?

doesn't matter now i guess as we can merge PRs with squashing.
I don't see the canges

SupaflyFPV:
> 1. set simplified_gyro_filter = OFF

ok forgot to push then made more changes...

I removed the filter freq to use the sliders only with simplified tuning apply. 

The tricky bit is the 42688 filter setup where I use the sliders for gyro and not dterm - let me know that is correct...

Hopefully all spacing is correct now...

limonspb:
> > 1. set simplified_gyro_filter = OFF
> 
> ok forgot to push then made more changes...
> 
> I removed the filter freq to use the sliders only with simplified tuning apply.
> 
> The tricky bit is the 42688 filter setup where I use the sliders for gyro and not dterm - let me know that is correct...
> 
> Hopefully all spacing is correct now...

Could you please open the files web view, and scroll carefully down, you will see the indentation is still off in many places:
https://github.com/betaflight/firmware-presets/pull/356/files
for example, your "default" filter settings got extra 4 spaces, they are not located in any option.
Also rates options.

And i still don't understand why you don't want to just apply simplified tuning at the end. It is not going to apply sliders that are OFF.

SupaflyFPV:
> 

ok done. Somehow missed them.

Ok I get it re sliders and simplified tuning command

limonspb (approved):

limonspb:
damn can't merge-and-squash
@blckmn could you please enable for this repo as well?

haslinghuis (dismissed):

limonspb (dismissed):
waiting for squash, or for enabling squash-merge

limonspb (approved):

PR #362, opened 2023-03-01
Create 1000hz_race.txt

sugaarK:
1000hz race preset... it does require the latest maintenance branch or 4.5 zulu right now though


limonspb (approved):
Preset lookgs good.
But you are in charge of the values :)

PR #379, opened 2023-04-05
update a few of my presets

sugaarK:
update a few of the karate presets and adding elrs to the keywords on the 1000hz preset

### DIFF
--- a/presets/4.3/rc_link/generic/1000hz_race.txt
+++ b/presets/4.3/rc_link/generic/1000hz_race.txt
@@ -3,7 +3,7 @@
 #$ FIRMWARE_VERSION: 4.4
 #$ CATEGORY: RC_LINK
 #$ STATUS: OFFICIAL
-#$ KEYWORDS: 1000Hz, 1ms, rc, rx, link, smoothing
+#$ KEYWORDS: 1000Hz, 1ms, rc, rx, link, smoothing, ELRS
 #$ AUTHOR: Sugar K
 #$ DESCRIPTION: Generic RC link settings for racing with 1000Hz RC link. Best used with 4.3.3+ or 4.4.1+
 #$ DESCRIPTION: WARNING: make ABSOLUTELY SURE that the OpenTx or EdgeTx Hardware ADC Filter is un-checked!
--- a/presets/4.4/tune/karate/karate_race.txt
+++ b/presets/4.4/tune/karate/karate_race.txt
@@ -1,7 +1,7 @@
 #$ TITLE: Karate Race 6s 5"
 #$ FIRMWARE_VERSION: 4.4
 #$ CATEGORY: TUNE
-#$ STATUS: EXPERIMENTAL
+#$ STATUS: OFFICIAL
 #$ KEYWORDS: karate, 6S, race, world champion, 5 inch, 5", sugarK, limon, ctzsnooze, KarateBrot
 #$ AUTHOR: sugarK
 
--- a/presets/4.4/tune/karate/karate_race.txt
+++ b/presets/4.4/tune/karate/karate_race.txt
@@ -35,8 +35,13 @@
 #$ DESCRIPTION:
 #$ DESCRIPTION: ## Second note... Radio links:
 #$ DESCRIPTION: 1. Make sure your radio firmware is up to date using either EdgeTX or OpenTX
-#$ DESCRIPTION: 1. Make sure your **ADC Filter is OFF** in the hardware page
-#$ DESCRIPTION: 1. Go to the radio (RC_LINK) presets and apply the correct setup for your system and link speed
+#$ DESCRIPTION: 2. Make sure your **ADC Filter is OFF** in the hardware page
+#$ DESCRIPTION: 3. Go to the radio (RC_LINK) presets and apply the correct setup for your system and link speed
+#$ DESCRIPTION:
+#$ DESCRIPTION: <br>
+#$ DESCRIPTION: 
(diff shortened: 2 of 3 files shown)


limonspb:
i had one guy recently complaining about weird wobbles on the latest karate preset. Here is his log:
[btfl_001.bbl.txt](https://github.com/betaflight/firmware-presets/files/11162033/btfl_001.bbl.txt)

wondering ig it could be because of the karate switch to sliders, but the pid values are somewhat off?

sugaarK:
> i had one guy recently complaining about weird wobbles on the latest karate preset. Here is his log: [btfl_001.bbl.txt](https://github.com/betaflight/firmware-presets/files/11162033/btfl_001.bbl.txt) (remove .txt)
> 
> wondering ig it could be because of the karate switch to sliders, but the pid values are somewhat off?
> 
> Also was thinking that dyn_notch_min_hz = 200 is too spicy for lots of drones out there. That's why we have changed the defaults to 100hz? And it some options you have it even at 250hz. It could be fine for Openracer/Viper, but think about other frame manufacturers lol

looked at the log its pretty clean and looks good  I agree on the dynamic the min should be 125 I think so It can catch any frame res.. other wise there isn't any thing bad in his log.. what's his PWM settings on the quad? I recommend 48khz locked pretty much always maybe it needs to be in the notes for the present what do you think @limonspb

limonspb:
> > i had one guy recently complaining about weird wobbles on the latest karate preset. Here is his log: [btfl_001.bbl.txt](https://github.com/betaflight/firmware-presets/files/11162033/btfl_001.bbl.txt) (remove .txt)
> > wondering ig it could be because of the karate switch to sliders, but the pid values are somewhat off?
> > Also was thinking that dyn_notch_min_hz = 200 is too spicy for lots of drones out there. That's why we have changed the defaults to 100hz? And it some options you have it even at 250hz. It could be fine for Openracer/Viper, but think about other frame manufacturers lol
> 
> looked at the log its pretty clean and looks good I agree on the dynamic the min should be 125 I think so It can catch any frame res.. other wise there isn't any thing bad in his log.. what's his PWM settings on the quad? I recommend 48khz locked pretty much always maybe it needs to be in the notes for the present what do you think @limonspb got more info from the pilot on the wobbles? im not seeing any thing obvious in the log

i think he was talking about these relatively slow wobbles: see its 30-40 degrees/sec off in some cases:
image

sugaarK:
ahh that's throbbles... bad CG would have a big impact on this as would using dynamic idle.. this is a case by case issue that a preset couldn't possibly  handle.. he could also try some thrust linear to try and help

limonspb:
> ahh that's throbbles... bad CG would have a big impact on this as would using dynamic idle.. this is a case by case issue that a preset couldn't possibly handle.. he could also try some thrust linear to try and help

ok cool.
i haven't tried myself the new version of the preset. And that was just a single feedback i had. So just making sure its ready to be official

limonspb:
> ahh that's throbbles... bad CG would have a big impact on this as would using dynamic idle.. this is a case by case issue that a preset couldn't possibly handle.. he could also try some thrust linear to try and help

ps: you sure you don't want to also lower `set dyn_notch_min_hz = 250` to something smaller? :)

sugaarK:
> > ahh that's throbbles... bad CG would have a big impact on this as would using dynamic idle.. this is a case by case issue that a preset couldn't possibly handle.. he could also try some thrust linear to try and help
> 
> ok cool. i haven't tried myself the new version of the preset. And that was just a single feedback i had. So just making sure its ready to be official

your not gona get throbbles if you don't zero the throttle so you might not get them at all..

sugaarK:
> > ahh that's throbbles... bad CG would have a big impact on this as would using dynamic idle.. this is a case by case issue that a preset couldn't possibly handle.. he could also try some thrust linear to try and help
> 
> ps: you sure you don't want to also lower `set dyn_notch_min_hz = 250` to something smaller? :)

that's for the spicy tune.. its not for every one but works fine with a lot of quads

limonspb:
> > > ahh that's throbbles... bad CG would have a big impact on this as would using dynamic idle.. this is a case by case issue that a preset couldn't possibly handle.. he could also try some thrust linear to try and help
> > 
> > 
> > ps: you sure you don't want to also lower `set dyn_notch_min_hz = 250` to something smaller? :)
> 
> that's for the spicy tune.. its not for every one but works fine with a lot of quads

i just not aware of any downsides of running lower min hz for dynamic notch. So it might cover more drones without much downsides if we lower it?

sugaarK:
200hz?

limonspb:
> 200hz?

you decide lol. I'm just asking questions to mess it all up :))

sugaarK:
the point of the spicy is to reduce filter delay... 200 I think is good

sugaarK:
done

limonspb:
> the point of the spicy is to reduce filter delay... 200 I think is good

as far as i understand, if you have it at 100Hz, its not going to increase delay unless your quad needs a notch at 120hz.
And if it needs 120Hz notch, its better have it than don't :)

sugaarK:
> > the point of the spicy is to reduce filter delay... 200 I think is good
> 
> as far as i understand, if you have it at 100Hz, its not going to increase delay unless your quad needs a notch at 120hz. And if it needs 120Hz notch, its better have it than don't :)

if you need notches at 120hz your setup isn't good enough to run the spicy tune.....

limonspb (approved):

PR #382, opened 2023-04-17
Fix firmware version in FPVian freestyle 5in tune

FPVian:
I did some testing with the new AG and decided to leave it at defaults. Defaults just keep getting better and better.


limonspb (approved):

FPVian:
@limonspb I just noticed this is marked as experimental even though it was ported to 4.4 from a 4.3 community preset. Does it need to be re-reviewed for each release?

limonspb:
> @limonspb I just noticed this is marked as experimental even though it was ported to 4.4 from a 4.3 community preset. Does it need to be re-reviewed for each release?

4.4 and 4.3 tunes are pretty much equal. I think now its safe to change it to community, if you have someone tested on 4.4

PR #383, opened 2023-04-19
Add QuadMcFly 5in Freestyle Tune

VirtualEnder:
Conservative 5inch Mid-weight Freestyle tune with GoPro

### DIFF
--- a/presets/4.4/tune/quadmcfly/QuadMcFly_5_Inch_Freestyle.txt
+++ b/presets/4.4/tune/quadmcfly/QuadMcFly_5_Inch_Freestyle.txt
@@ -1,0 +1,70 @@
+#$ TITLE: QuadMcFly - 5" Freestyle with GoPro (Mid-Weight)
+#$ FIRMWARE_VERSION: 4.4
+#$ CATEGORY: TUNE
+#$ STATUS: EXPERIMENTAL
+#$ KEYWORDS: 4S, 6S, 5in, 5", freestyle
+#$ AUTHOR: QuadMcFly (Ryan Harrell)
+
+#$ PARSER: MARKED
+
+#$ DESCRIPTION: QuadMcFly 5" Freestyle Tune (Mid Weight)
+#$ DESCRIPTION: -----------
+#$ DESCRIPTION:
+#$ DESCRIPTION: <a href="https://www.youtube.com/c/QuadMcFlyFlies" target="_blank"><img src="https://www.miniquadtestbench.com/assets/images/me-small-dji.png" width="125px" style="float:left; margin-right: 25px; display: block;"/></a><br />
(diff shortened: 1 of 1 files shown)


VirtualEnder:
Added URL for PR

haslinghuis (approved):

limonspb (changes requested):
.

presets/4.4/tune/quadmcfly/QuadMcFly_5_Inch_Freestyle.txt
```diff
@@ -0,0 +1,70 @@
...
+set pidsum_limit = 1000
+set pidsum_limit_yaw = 1000
+
+set dyn_idle_min_rpm = 30
+set thrust_linear = 40
+set throttle_boost = 15
+set tpa_mode = PD
+set iterm_relax_cutoff = 12
+set vbat_sag_compensation = 100
+set small_angle = 180
```
limonspb, line 62: this should not be in the tune
it could be, but then in the UNCHECKED option

presets/4.4/tune/quadmcfly/QuadMcFly_5_Inch_Freestyle.txt
```diff
@@ -0,0 +1,70 @@
...
+
+# -- QuadMcFly Filter Settings --
+
+set dshot_bidir = ON
```
limonspb, line 66: plz set also dshot 600, prior to setting dshot_bidir .

limonspb (changes requested):

presets/4.4/tune/quadmcfly/QuadMcFly_5_Inch_Freestyle.txt
VirtualEnder: I can see this argument for the vbat compensation and I can probably just remove the small angle. However I would argue that the PD for TPA and the I-term relax cutoff are integral parts of the tune itself. I would have to retune the core gain sliders if I changed those values. Primarily allowing the TPA option to be unchecked would cause the p gains to need to be lowered as well when it was unchecked, and allowing I term relax cutoff to be unchecked would mean the I gains would need to be adjusted separately as well. I could do this, but it would require a significant tune adjustment beyond just putting these in as options, which I feel isn't the goal of this tune.

presets/4.4/tune/quadmcfly/QuadMcFly_5_Inch_Freestyle.txt
VirtualEnder: Yeah I realized that last night, hah, I'll adjust that shortly.

VirtualEnder:
I rebased the code, and added the changes, but now it won't let me reopen the pull request. Should I open a new PR?

limonspb:
you didn't have to close PR or remove the remote branch. But its ok

VirtualEnder:
Yeah I accidentally did a hard reset and force push in the wrong branch and I lost an intermediate commit, so I couldn't reopen it :/ Oh well.

PR #387, opened 2023-04-20
Add QuadMcFly 5in Freestyle Tune

VirtualEnder:
Mid-weight freestyle build with options for Dshot 600 and Dshot 300. RPM Filtering required. Update from broken pull request #383

### DIFF
--- a/presets/4.4/tune/quadmcfly/QuadMcFly_5_Inch_Freestyle.txt
+++ b/presets/4.4/tune/quadmcfly/QuadMcFly_5_Inch_Freestyle.txt
@@ -1,0 +1,86 @@
+#$ TITLE: QuadMcFly - 5" Freestyle with GoPro (Mid-Weight)
+#$ FIRMWARE_VERSION: 4.4
+#$ CATEGORY: TUNE
+#$ STATUS: EXPERIMENTAL
+#$ KEYWORDS: 4S, 6S, 5in, 5", freestyle
+#$ AUTHOR: QuadMcFly (Ryan Harrell)
+
+#$ PARSER: MARKED
+
+#$ DESCRIPTION: QuadMcFly 5" Freestyle Tune (Mid Weight)
+#$ DESCRIPTION: -----------
+#$ DESCRIPTION:
+#$ DESCRIPTION: <a href="https://www.youtube.com/c/QuadMcFlyFlies" target="_blank"><img src="https://www.miniquadtestbench.com/assets/images/me-small-dji.png" width="125px" style="float:left; margin-right: 25px; display: block;"/></a><br />
+#$ DESCRIPTION: After quite a while refining a base tune across multiple 5 inch setups I think I've gotten a very robust starting point that is worth sharing. This tune is designed to be somewhat conservative to handle things like dinged props and imperfect motors, but still handle well enough to minimize prop wash and feel sharp and connected.
+#$ DESCRIPTION: <br /><br /><br />
+#$ DESCRIPTION: 
+#$ DESCRIPTION: This should handle anything from around 650g up to close to 800g. Above 750g it might start feeling sluggish, but it should be a good starting point for any mid-weight 5 inch build. Only small adjustments to the P and D sliders should be necessary for most builds. This tune works well for me on both F4 and F7 processors at 8khz and Dshot 600. If you have a noisy ESC or ESCs with poorly designed input filtering you may need to drop to DShot 300, but otherwise it should be fine.
+#$ DESCRIPTION: <br /><br />
+#$ DESCRIPTION: <strong>This tune requires ESCs capable of RPM telemetry (BLheli 32, AM32, or BlueJay/JESC).</strong>
+#$ DESCRIPTION: 
+#$ DESCRIPTION: Suggested ESC settings are:
+#$ DESCRIPTION: -----------
+#$ DESCRIPTION:  - Timing: 23 degrees
+#$ DESCRIPTION:  - PWM: 48khz fixed (May also work with 24khz and ByRPM)
+#$ DESCRIPTION:
+#$ DESCRIPTION: Suggested Propellers:
+#$ DESCRIPTION: -----------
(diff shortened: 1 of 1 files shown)


presets/4.4/tune/quadmcfly/QuadMcFly_5_Inch_Freestyle.txt
```diff
@@ -0,0 +1,99 @@
...
+#$ OPTION_GROUP BEGIN: Motor Protocol Selection (Choose One)
+
+    #$ OPTION BEGIN (CHECKED): DSHOT 600
+        set motor_pwm_protocol = DSHOT600
+        set dshot_bidir = ON
+        # -- QuadMcFly Filter Settings --
+        set simplified_gyro_filter_multiplier = 200
+        set simplified_dterm_filter_multiplier = 120
+
+        simplified_tuning apply
```
limonspb, line 74: this can be done just once at the end of the preset

presets/4.4/tune/quadmcfly/QuadMcFly_5_Inch_Freestyle.txt
```diff
@@ -0,0 +1,99 @@
...
+set throttle_boost = 15
+set tpa_mode = PD
+set iterm_relax_cutoff = 12
+
+# ------ OPTIONS GO BELOW THIS LINE ------
+# This is where the author includes options that require input from the User
+
+#$ OPTION_GROUP BEGIN: Motor Protocol Selection (Choose One)
+
+    #$ OPTION BEGIN (CHECKED): DSHOT 600
```
limonspb, line 67: these two options are a little misleading. Besides setting up dshot, they also set some filter values despite it says "motor protocol"

My suggestion:

```
set ..... all your tuning params
set simplified_gyro_filter_multiplier = 200
set simplified_dterm_filter_multiplier = 120
set motor_pwm_protocol = DSHOT600
set dshot_bidir = ON

#$ OPTION BEGIN (UNCHECKED): DSHOT 300
    set motor_pwm_protocol = DSHOT300
#$ OPTION END

.....
...

at the very end:
simplified_tuning apply
```
so by default, it will set bidirectional dshot 600. But if someone needs, he can select 300 with the checkbox
What do you think?

limonspb (approved):

limonspb (changes requested):
left comments in the code above ^

presets/4.4/tune/quadmcfly/QuadMcFly_5_Inch_Freestyle.txt
```diff
@@ -0,0 +1,99 @@
...
+#$ DESCRIPTION:  - Gemfan 51466 Hurricane
+#$ DESCRIPTION:
+#$ DESCRIPTION:
+#$ DESCRIPTION: Suggested Motors:
+#$ DESCRIPTION: -----------
+#$ DESCRIPTION:  - ~2500kv on 4S
+#$ DESCRIPTION:  - ~1700kv on 6S
+#$ DESCRIPTION:
+#$ DESCRIPTION:
+#$ WARNING: Requires RPM filtering, and enables Dynamic Idle and Thrust Linear. I run fairly high Thrust Linear, so low throttle may be more agressive than some other tunes.
```
limonspb, line 41: Can also add:
```
#$ INCLUDE_WARNING: misc/warnings/en/dshot.txt
#$ INCLUDE_WARNING: misc/warnings/en/rpm_filters.txt
```

presets/4.4/tune/quadmcfly/QuadMcFly_5_Inch_Freestyle.txt
VirtualEnder: Cool, yeah that works. I wasn't sure on if anything was allowed to happen after the option settings. Will update.

presets/4.4/tune/quadmcfly/QuadMcFly_5_Inch_Freestyle.txt
VirtualEnder: Moved

VirtualEnder:
Ok, I think those issues are all resolved, let me know if there is anything else. I've tested locally and it all seems to be working as expected.

limonspb:
> Ok, I think those issues are all resolved, let me know if there is anything else. I've tested locally and it all seems to be working as expected.

great job!
I think the only question i have, why you calling dshot 300 a performance boost?
Dshot 300 is really for those having electrically noisy drones, that can't handle dshot 600.

VirtualEnder:
> 

I usually see it recommended for F4 processors to save CPU cycles in combination with slower loop time. I was actually considering putting the the code for the slower loop time in there as well. I can change that if you think that's not necessary. 8khz and DShot 600 seems to work for me on F4, but it's pretty high CPU with no other features enabled.

limonspb:
> > 
> 
> I usually see it recommended for F4 processors to save CPU cycles in combination with slower loop time. I was actually considering putting the the code for the slower loop time in there as well. I can change that if you think that's not necessary. 8khz and DShot 600 seems to work for me on F4, but it's pretty high CPU with no other features enabled.

yeah, on F4 CPUs its better to run 4k. But it combines pretty good with dshot 600.
Dshot 300 really only needed when the FC->ESC connection is noisy or FC/ESC design is bad.

VirtualEnder:
> > > 
> > 
> > 
> > I usually see it recommended for F4 processors to save CPU cycles in combination with slower loop time. I was actually considering putting the the code for the slower loop time in there as well. I can change that if you think that's not necessary. 8khz and DShot 600 seems to work for me on F4, but it's pretty high CPU with no other features enabled.
> 
> yeah, on F4 CPUs its better to run 4k. But it combines pretty good with dshot 600. Dshot 300 really only needed when the FC->ESC connection is noisy or FC/ESC design is bad.

Cool, I'll change that wording then.

VirtualEnder:
Ok, I updated the description and option to reflect what we discussed. Let me know if there is anything else.

limonspb (approved):

PR #391, opened 2023-05-02
Add FrSky R9 RC link preset (FPORT, 6.67 ms mode)

Jakub-Espandr:
Due to missing FrSky R9 preset, I created RC LINK preset with modified setpoint and feedforward filtering.
I've included port selection, a choice connection the RX to the FC (Serial_RX_inversion), and various filtering options depending on intended flight style.

### DIFF
--- a/presets/4.4/rc_link/FrSky R9 (6.67ms).txt
+++ b/presets/4.4/rc_link/FrSky R9 (6.67ms).txt
@@ -1,0 +1,131 @@
+#$ TITLE: FrSky R9 (6.67 ms)
+#$ FIRMWARE_VERSION: 4.3
+#$ FIRMWARE_VERSION: 4.4
+#$ CATEGORY: RC_LINK
+#$ STATUS: EXPERIMENTAL
+#$ KEYWORDS: FrSky, R9, 6.67 ms, ACCESS, FPORT, rc, rx , link
+#$ AUTHOR: Jakub Espandr (FlyCamCzech)
+#$ DESCRIPTION: RC link settings for a 150 Hz (6.67 ms PWM) Frsky R9 FLEX ACCESS.
+
+#$ DESCRIPTION: WARNING: Make ABSOLUTELY SURE that the Hardware ADC Filter is unchecked in OpenTx/EdgeTx!
+#$ DESCRIPTION: WARNING: Check that you have selected 6.67 ms PWM mode in the model settings in the receiver tab (under Modules and Preferences)!
+#$ DESCRIPTION: WARNING: All cinematic settings are very smooth, resulting in a noticeable delay in stick response.
+#$ DESCRIPTION: WARNING: Frsky R9 module does not have a consistent packet rate, so 4-point FF averaging is required!
+#$ DESCRIPTION:
+#$ DESCRIPTION:
+#$ DESCRIPTION: * Modern FrSky receivers use the FPORT protocol, which requires only a single cable that acts as RC command output from the receiver and telemetry input from the FC.
+#$ DESCRIPTION: * Due to the fact that FPORT is an uninverted protocol, when connecting the SPORT wire from RX to TX pin on the FC, you must choose the correct preset.
+#$ DESCRIPTION:
+#$ DESCRIPTION:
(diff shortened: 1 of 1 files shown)


Jakub-Espandr:
> Hi
> 
> you mind sharing how did you figured the values for racing/freestyle etc?


Hello,

**Racing** with ff jitter factor = 8 and rc_smoothing_auto_factor = 35 is based on Betaflight's recommended value of 4-5 for jitter factor. I tested these values on different setups with FrSky R9M, R9mini, R9 Slim OTA + receivers and the signal is still sometimes crappy, so I tried playing with the jitter factor and found that a value of 8 is quite clean with not so significant delay added to the raw commands. From the Betaflight tuning notes, jitter factor = 8 should add 0.8% RC command attenuation per step. The default RC smoothing autofactor is 30 and from my tests at this value there were some uneven steps that were transferred through the setpoint to the P-term.

**Freestyle** with feedforward_jitter_factor = 9 produced fairly clean traces with no noise from uneven packets. The Betaflight tuning notes suggested rc_smoothing_auto_factor 60 for HD freestyle and the default is 30, so I tried playing with different values in increments of 5 and ended up with 40, which did not add any significant amount of delay between the raw RC command and the setpoint. Still with only these settings there were some steps with equal intervals in setpoint so I tried playing with rc_smoothing_setpoint_cutoff and feedforward cutoff frequency to not create huge amount of delay. I ended up with 25 Hz, which produced a fairly smooth setpoint suitable for freestyle flying. From 10 flight and logs there are sometimes some micro steps in the traces, but should definitely be OK for this type of flight.

**HD Freestyle** has a slightly higher value than suggested in the BF notes. It has a value of 80 compared to the recommended 60, but when I compared HD footage from the DJI O3 and bladed it to the flight log I saw some micro movements that were exactly in the same place as micro steps on the setpoint and P-term traces. With this value of 80, these microsteps disappeared. Also from tuning notes with manual filtering of the RC link, it was written that a smaller number will produce cleaner and smoother inputs. I tested several flights with 5" (Titan DC5 with O3) and 4" (custom Explorer LR4 O3) and found that a cut-off frequency of 20 Hz produces a perfect balance of smooth setpoint and delay.

**Cinematic preset** Although the feedforward smoothness factor should vary with the data rate, I set it to 35 and the jitter factor to 12, as recommended for cinematic flights. When I compared ff_smooth_factor 30 against 35 and 40 with JitterFactor 12, it turned out that the smoothest value of 35 produced slightly cleaner traces on the P-term, as the feedforward was filtered more. I tested it on extreme situations with feedforward twice as large as the optimal values I got from the basement pidtuning method. Rc_smoothing_auto_factor was suggested to be around 120 for pure cinematic flying. I almost always fly long range cinematic cruises and on longer ranges my R9 system produced inconsistent frame delta and frame age when I flew behind an obstacle and the RSSI dropped. This caused marginal differences in these two values with some micro-steps in Setpoint and P-Term recorded with BlackBox and raw 4K footage. I tested it several times in fields, woods and super low flying with RSSI around 40. I think a lot of cinematic pilots have quite a lot of expo in their rates, so if it compensates for the smoothness of the signal while picking up the delay between RX and Setpint due to the higher smoothing factor, it shouldn't be a detriment as this style of flying strives for perfectly clean movements and doesn't mind the longer response time.

The **Super Cinematic** preset is an extreme combination with a significant amount of delay caused by RC link filtering, but it produces super clean setpoint traces with absolutely no spikes or steps. It has an RC smoothing auto factor of 250, which is twice the recommended BF and produces twice the delay. The more I fly this preset with a 4" O3 setup, the more I like it.
It has a huge amount of filtering on the RPY axes and a lot of filtering on the throttle too. With this amount of filtering I was able to get smooth setpoint traces even with micro-shakes on the RC sticks caused by frozen fingers during cold winter flights. The RC_smoothing_setpoint_cutoff frequency was reduced to 6Hz, which ultimately smoothed out any movement, which can help to get smooth footage from the onboard HD camera. From my tests, this preset works well with O3 AirUnit footage stabilised in GyroFlow or even with onboard RockSteady.

limonspb (approved):

PR #401, opened 2023-06-13
adding 12 motor poles to whoop tunes

limonspb:
i think it's safe to set 12 poles for whoops no matter what.

whoop guys started running RPM spec, and that becomes critical.


haslinghuis (approved):

sugaarK:
You missed karate whoop

PR #402, opened 2023-06-18
Karate whoop 4.4, added 12 motor poles

limonspb:
missed one in https://github.com/betaflight/firmware-presets/pull/401


sugaarK (approved):
Thanks @limonspb

PR #413, opened 2023-09-09
Add TUNE to OTHER category

antoni9:
Hi,

I created this tune as part of my frame bundle [[100-euro-fpv.com](url)] and now sharing it as a preset for BFC.

It contains multiple category changes - a full setup from a blank 4.4 flash, hence why it falls under the OTHER category.

Thanks for looking into this
Tony

### DIFF
--- a/presets/4.4/other/100-euro-fpv-tune.txt
+++ b/presets/4.4/other/100-euro-fpv-tune.txt
@@ -149,7 +149,7 @@ set thrust_linear = 25
 #$ OPTION END
 
 #$ OPTION BEGIN (CHECKED): Min/max check 1000/2000
-    set min_check = 1000
+    set min_check = 1010
     set max_check = 2000
 #$ OPTION END
 
--- a/presets/4.4/other/100-euro-fpv-tune.txt
+++ b/presets/4.4/other/100-euro-fpv-tune.txt
@@ -148,7 +148,7 @@ set thrust_linear = 25
     set dshot_bidir = ON
 #$ OPTION END
 
-#$ OPTION BEGIN (CHECKED): Min/max check 1000/2000
+#$ OPTION BEGIN (CHECKED): Min/max check 1010/2000
     set min_check = 1010
     set max_check = 2000
 #$ OPTION END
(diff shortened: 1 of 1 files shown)


limonspb (approved):

limonspb:
Looking good. Surprizing to see that as a first PR to be clean.
minor question about this:
```
    set min_check = 1000
    set max_check = 2000

```as far as i understand it only affects the stick commands.

haslinghuis (approved):

antoni9:
@limonspb Thanks! The min/max checks increase perceived stick resolution, but you are right the min_ one should be >1000 otherwise BF will not ARM.

haslinghuis (approved):

limonspb:
> @limonspb Thanks! The min/max checks increase perceived stick resolution, but you are right the min_ one should be >1000 otherwise BF will not ARM.

i don't think it increases stick resolution. I was thinking the same before. But dug a little in the code and saw it's being used only for stick commands.

antoni9:
To expand on the "BF will not arm" statement - it will not arm because min_check is equal to the throttle min value of 1000 being received from a well calibrated radio, if the radio is calibrated to 998 there's no issue.

limonspb:
> To expand on the "BF will not arm" statement - it will not arm because min_check is equal to the throttle min value of 1000 being received from a well calibrated radio, if the radio is calibrated to 998 there's no issue.

also for max check=2000, do the stick commands still work, if the radio is calibrated up to 1999?

antoni9:
afaik, only min_check on throttle prevents arming, on other axes this only contributes to a tiny misalignment in the endpoint and presumably through the entire curve to compensate for the 1 - 2 ms difference. I have my radio outputting 998 - 2002

limonspb (approved):

limonspb:
> afaik, only min_check on throttle prevents arming, on other axes this only contributes to a tiny misalignment in the endpoint and presumably through the entire curve to compensate for the 1 - 2 ms difference. I have my radio outputting 998 - 2002 for symmetry.

What i'm worried about, is that MAX_CHECK=2000 will block stick commands (like OSD menu, profile changes, etc) in case radio calibrated that way that it's not giving 2000, but lets say 1999

PR #417, opened 2023-10-30
disable colored warnings for HDZERO Preset in 4.5

MacDaddyFPV:
Created a BF 4.5 Version of the HDZERO VTX preset to add in the disable colored warnings as they are currently not compatible with HDZERO.
Tested from my fork as a source.


MacDaddyFPV:
Discussed with SugarK and decided adding into the current preset is a better method

sugaarK:
> Discussed with SugarK and decided adding into the current preset is a better method

Changed my mind! This is better @MacDaddyFPV

sugaarK (approved):

PR #430, opened 2023-12-08
4.5 tune defaults changes

limonspb:
Continuation of this PR:
https://github.com/betaflight/firmware-presets/pull/429

few params are renamed/added in comparisson with 4.4 tune defaults.

I also made this list in the same order it appears in CLI for easier maintenance in the future.

Plus i have removed these two parameters as they probably don't belong to tuning:
```
vbat_warning_cell_voltage
motor_pwm_rate
```

these three parameters might be renamed:
```
set tpa_rate_lower = 20
set tpa_breakpoint_lower = 1050
set tpa_breakpoint_lower_fade = ON
```
if this PR is merged:
https://github.com/betaflight/betaflight/pull/13206
but thats an easy fix

### DIFF
--- a/presets/4.5/tune/defaults.txt
+++ b/presets/4.5/tune/defaults.txt
@@ -4,12 +4,11 @@
 #$ STATUS: OFFICIAL
 #$ KEYWORDS: defaults, reset, tune, pid, basic, default
 #$ AUTHOR: Betaflight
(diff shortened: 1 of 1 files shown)


ctzsnooze (approved):
Lots of work there - thank you!

haslinghuis (approved):

PR #435, opened 2023-12-11
Other presets 4.4->4.5

limonspb:
Who is making the most complicated presets?
Or course it's @TehllamaFPV  :)

Moved three presets 4.4/other to 4.5/other, and replaced defaults with the correct ones.


haslinghuis (approved):

TehllamaFPV:
Well... somehow I don't think I made that worse

PR #449, opened 2024-03-24
add vtx table preset JHEMCU_VTX20-600

jikanos:
VTX table received from manufacturers directly


limonspb (approved):

limonspb:
Hi!
Whats the difference between the same file here?
https://github.com/betaflight/firmware-presets/pull/450

PR #454, opened 2024-04-06
Created UWL Whoop tune by FreshBread

Stampede10343:
This is a Underground Whoop League (UWL) tune  created by FreshBread.

### DIFF
--- a/presets/4.5/tune/UWL-FreshBread.txt
+++ b/presets/4.5/tune/UWL-FreshBread.txt
@@ -1,0 +1,69 @@
+#$ TITLE: UWL Tune by FreshBread
+#$ FIRMWARE_VERSION: 4.5
+#$ CATEGORY: TUNE
+#$ STATUS: EXPERIMENTAL
+#$ KEYWORDS: UWL, underground, Whoop, 75mm, tune, acro, 1S, FreshBread
+#$ AUTHOR: FPV_Cam
+#$ DESCRIPTION: A tune for Underground Whoop League 75mm 1S whoop racing by FreshBread
+#$ WARNING: You will need to ensure your ESC firmware supports Bidirectional DSHOT for the RPM limiter
+#$ DISCUSSION: https://github.com/betaflight/firmware-presets/pull/nn
+#$ FORCE_OPTIONS_REVIEW: true
+
+#$ INCLUDE: presets/4.5/tune/defaults.txt
+
+# Tune begin
+
+set gyro_lpf1_static_hz = 0
+set gyro_lpf2_static_hz = 0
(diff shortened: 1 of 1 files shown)


limonspb (changes requested):
sorry just saw the PR.
Could you plz add filter defaults include.
Also you don't have to type all PIDs/Filter values, if you are using sliders.
You can just list slider values that are different from defaults
And add
"simplified_tuning apply" at the end.
Thanks!

limonspb:
> @limonspb I believe I've got it fixed up, take a look when you get a chance, thanks!

Sweet thanks!
Few more notes if you don't mind:
Title should prolly include "whoop 75mm" so people could see right away it's for whoops. 

In the description I see it's mentioning kaack. Since this is 4.5 preset, you don't need kaack, can just flash with RACE_PRO option selected from the drop-down of options.

BMI option better be unchecked prolly?

Stampede10343:
> > @limonspb I believe I've got it fixed up, take a look when you get a chance, thanks!
> 
> Sweet thanks! Few more notes if you don't mind: Title should prolly include "whoop 75mm" so people could see right away it's for whoops.
> 
> In the description I see it's mentioning kaack. Since this is 4.5 preset, you don't need kaack, can just flash with RACE_PRO option selected from the drop-down of options.
> 
> RPM limit better be in a separate option from setting up filters, and RPM limit option needs to be unchecked by default. The RPM filters setting, motor poles etc can be without option, if you want.
> 
> BMI option better be unchecked prolly?

Sounds good. I was wondering how to enable it in 4.5, good to know about that option!

limonspb:
was wondering about this:

#$ OPTION BEGIN (UNCHECKED): UWL RPM Limiter Settings
    set dshot_bidir = ON
    set motor_poles = 12
    set rpm_limit = ON
    set rpm_limit_value = 31500
#$ OPTION END

basically it means if someone doesn't want to apply the spec limit, he also won't have bidirectional dshot?

Stampede10343:
> was wondering about this:
> 
> #$ OPTION BEGIN (UNCHECKED): UWL RPM Limiter Settings set dshot_bidir = ON set motor_poles = 12 set rpm_limit = ON set rpm_limit_value = 31500 #$ OPTION END
> 
> basically it means if someone doesn't want to apply the spec limit, he also won't have bidirectional dshot?

I think the main focus is the UWL spec, if someone wants RPM filtering I feel like its fine to let them do that on their own. I'm unsure how well this tune would work on a lightweight or more 'standard' 75mm so I didn't focus too much on the non UWL whoops

limonspb (approved):

PR #460, opened 2024-05-05
Added OpenRacer preset by JazzMutant

Jaz7Mutant:
Created preset for OpenRacer 6S 5 inch build


limonspb:
Needs few minor tweaks:
1. Sag compensation can be an option, but must be unchecked by default.
2. DShot 600 and bidirectional DShot could be on without option, and just a separate option with DShot 300.
3. Simplified tuning apply can be just once at the very end.
4. Normal filter option must be selected by default? Not sure, you decide.

limonspb (approved):

haslinghuis (approved):

Memasukov:
Set up cool and safe. I have already configured three open racers on this preset.

BurundukFPV:
Just as a FYI, if running bluejay motor braking the quad will freak out in low throttle maneuvers with this preset.

PR #464, opened 2024-05-09
Aos Presets

ChrisRosser:
Hi Everyone,

I have just updated my presets for BF4.5 and was wondering if it would be appropriate to merge them into the main presets repo?

Lots of pilots are using them from my repo on BF 4.3 4.4 and now 4.5 so they should be very safe to merge.

Let me know if its appropriate or not!

Thanks,

Chris


limonspb:
sould be ok. Its a lot, but they seem to be good. Ill go through them this weekend more carefully, see if i can spot anything to fix.

limonspb:
went thought, have a few minor things:

Options like down below can't be checked by default inside the TUNE/FILTERS because these parameters technically personal preferences, and not the tunes.

SO you can have them as options, but should not be applied by default. Only when user mindfully clicks on them:

```
#$ OPTION_GROUP BEGIN: Check all of these (recommended)

    # -- VBat sag compensation --
    #$ OPTION BEGIN (CHECKED): Full battery sag compensation
        set vbat_sag_compensation = 100
        set vbat_sag_lpf_period = 2
    #$ OPTION END

    # -- Stick behaviour --
    #$ OPTION BEGIN (CHECKED): No stick deadband
        set deadband = 0
        set yaw_deadband = 0
    #$ OPTION END

    # -- Startup and arming --
    #$ OPTION BEGIN (CHECKED): Arm at any angle
        set small_angle = 180
    #$ OPTION END

    # -- Prop direction --
    #$ OPTION BEGIN (CHECKED): Props out (check motor direction!)
        set yaw_motors_reversed = ON
    #$ OPTION END

#$ OPTION_GROUP END
```

ChrisRosser:
@limonspb No problem I will set all those options as default unchecked 👍

ChrisRosser:
@limonspb I've made those changes as requested.

limonspb (approved):

PR #476, opened 2024-07-11
broke out throttle curves from rates and added secret sauce for stick feelz…

sugaarK:
updated karates to break out throttle curve from rates file to suit build types, added radio presets to TT tune, add secret sauce stick feels and warning for overclocking using bmi gyro


haslinghuis (approved):

limonspb (approved):

KarateBrot (approved):

limonspb:
Id rename secret sauce to ff jitter reduction off. Just so that people understood what they are clicking on

PR #486, opened 2024-08-30
HDZero VTX Table update

nerdCopter:
- https://discord.com/channels/868013470023548938/930285779077459968/1256901718537408532
- untested.
- note: HDZero sends tables from MSP overwriting any user config.
- note: HDZero's table uses the text `CUSTOM` not `FACTORY`. unsure of any significance of such.

### DIFF
--- a/presets/4.5/vtx/HDZero.txt
+++ b/presets/4.5/vtx/HDZero.txt
@@ -33,12 +33,12 @@
 # vtxtable
 vtxtable bands 6
 vtxtable channels 8
-vtxtable band 1 BOSCAM_A A FACTORY    0    0    0    0    0    0    0    0
-vtxtable band 2 BOSCAM_B B FACTORY    0    0    0    0    0    0    0    0
-vtxtable band 3 BOSCAM_E E FACTORY    0    0    0    0    0    0    0    0
-vtxtable band 4 FATSHARK F FACTORY    0 5760    0 5800    0    0    0    0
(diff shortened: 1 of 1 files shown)


limonspb:
@sugaarK could you plz check

sugaarK:
The original preset is the legal race band channels for most of the world and what HDzero published. hDzero added  some channels that are not legal to the VTX ( even more of you unlock ) and MSP  VTX automatically overwrites the VTX table with its current firmware VTX table on power up and handshake with the FC. So it’s non issue but I’m not sure we change the preset as it’s as legal every where as the one..

sugaarK:
@ctzsnooze

ctzsnooze:
@sugaarK - check the proposed change in the commit.  

The channels in the modified HDZero list, which get flashed to the FC, are:
- RaceBand 1-8 (5658-59217)
- Fatshark 1, 2 and 4 (5740-5800)
- Boscam E 1 (5705)

I'm sure some of these are not legal, but the outliers are the RaceBand extremes, and perhaps Boscam E1, which is a bit low.  The Fatshark ones are in the middle of the standard FPV frequency ranges.

The values in our Preset won't work unless there is telemetry, to let the HDZero goggles can over-ride our Preset.  Having a Present that won't work makes our Preset a bit pointless.

ctzsnooze (approved):

haslinghuis (approved):

sugaarK (approved):

PR #518, opened 2025-09-30
ozr_5x.txt tune - added ICM option as requested by author

limonspb:
Added ICM option to ozr_5x.txt as requested by DinDrones

### DIFF
--- a/presets/4.5/tune/ozr_5x.txt
+++ b/presets/4.5/tune/ozr_5x.txt
@@ -4,6 +4,7 @@
 #$ STATUS: EXPERIMENTAL
 #$ KEYWORDS: ozr, dindrones, 6s, 5 inch, 5", racing, 5in, din
 #$ AUTHOR: Benjamin Freelove
+#$ FORCE_OPTIONS_REVIEW: TRUE
 
 #$ PARSER: MARKED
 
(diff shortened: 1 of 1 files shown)


limonspb:
> frankly if your going to turn off an LPF it should be number 1 and leave number2 on.

we gonna leave it to authors at their own risk :)

presets/4.5/tune/ozr_5x.txt
```diff
@@ -43,6 +44,16 @@ set dyn_notch_max_hz = 650
 # -- Dterm filtering --
 set dterm_lpf1_dyn_expo = 7
 
+#$ OPTION_GROUP BEGIN: (EXCLUSIVE) Gyro type
+    #$ OPTION BEGIN (CHECKED): MPU6000
+        # empty on purpose, only ICM needs corrections
+    #$ OPTION END
+    #$ OPTION BEGIN (UNCHECKED): ICM42688
+        set gyro_lpf2_static_hz = 0
```
haslinghuis, line 52:
```suggestion
        set gyro_lpf1_static_hz = 0
```

KarateBrot: Adding to this: Gyro LPF2 is acting as an anti-alias filter. LPF1 is not. So if you want to disable a gyro lowpass, disable Gyro LPF1, just as @haslinghuis suggested. You get some benefits by running LPF2 instead of LPF1.

XaocuHKa: ok, let me run it this weekend! will update asap

sugaarK:
> Should be good with the second one on.  That's what the regular tune has and I bet it works also and is more conservative. Lets risk it!

The second lpf includes some anti aliasing depending on your pid rate selected. Also if you have tunes for the upcoming release its time to start thinking about new tunes for that

presets/4.5/tune/ozr_5x.txt
XaocuHKa: > 

I have reviewed the preset and gyro_lpf1_static_hz = 0 is already present in the original file. So the proposed change is to only remove gyro_lpf2_static_hz = 0, which we dont want as it was intentional and ICM gyros fly better with Gyro LPF2 off. 

Please approve as submitted originally.

PR #585, opened 2026-07-07
Rc link preset   2026.6 compatibility

Canaill51:
Hi guys,
This PR request is to update current RCLink presets.
Knowing that BF 2026.6 is now entered to RC phase, I thought it could be helpful to update RCLink preset according to.

Please find my PR and let me know if this needs additional changes.
Thanks in advance.

## Summary by CodeRabbit

* **New Features**
  * Added support for firmware version **2026.6**.
  * Updated RC link presets so they’re available for both **2025.12** and **2026.6**.
* **Bug Fixes**
  * Refreshed firmware version metadata across multiple preset profiles to improve configuration matching.
* **Chores**
  * Updated the stored package/index hash to reflect the latest preset changes.

PR #504, opened 2025-03-25
AOS UL8 Filter and Tune

ChrisRosser:
Hi Betaflight team!

Another day another preset. This one is for the AOS UL8 frame. The same format as previous submissions.

Please let me know if any changes are needed.

PR #494, opened 2024-11-01
Presets for new AOS Cine Frames

ChrisRosser:
Hi Everyone,

I'm launching some new Cine frames over the next few days and wanted to create presets so that people find them easy to tune. The tunes should be in the same style as previous submissions.

Thank you in advance for taking a look over them and I stand ready to make any alterations needed to esnure they meet our high Betaflight standards!

Also if you have any suggestions to improve the tunes based on your experience that would be incredible also.

PR #592, opened 2026-08-13
Add 2026.6 support and AOS 6 presets, consolidate the AOS filter presets

ChrisRosser:
Now that 2026.6 is out, this declares it on all of my AOS presets, in the same way as #585 does for the RC_LINK presets, adds presets for the AOS 6, and removes a set of duplicate filter presets while doing it.

### Firmware version declarations

- `2026.6` added to the tunes in `presets/2025.12/tune/aos_rc/`.
- `presets/4.5/rates/AOS_rates.txt` gains both `2025.12` and `2026.6`. It was never ported to the 2025.12 folder, so it is currently invisible to users on either recent release. All of its parameters are still valid, so nothing else needed changing.

`2026.6` rather than `2026.06`, since `FC_VERSION_STRING` is built from the unpadded month and Configurator matches with `flightControllerVersion.startsWith(firmware_version)`.

### New AOS 6 presets

`aos_6_filters` (4.5 / 2025.12 / 2026.6) and `aos_6_tune` (one file for 4.5, one covering 2025.12 and 2026.6). The tune values are carried over from the AOS 5.5 presets, with the AOS 6 reference build (1500mAh 6S, 1570KV) in the description.

### Consolidating the filter presets

The 19 filter presets under `presets/2025.12/filters/aos_rc/` were identical to their `presets/4.5` counterparts apart from the version line and an `OPIONS` -> `OPTIONS` typo. Rather than maintain two near-identical copies and add a third for 2026.6, the `presets/4.5` files now declare `4.5`, `2025.12` and `2026.6`, and the 2025.12 duplicates are removed. Net effect is 19 fewer files with no change to what gets applied.

Every CLI setting used by these filter presets is valid in all three releases, which is what makes a single file workable. `presets/4.5/filters/defaults.txt`, which they all include, gains `2025.12` and `2026.6` for the same reason â€” it is the only FILTERS defaults preset in the repo, and its values still match the 2026.6 firmware defaults (gyro 250/500, dterm 75/150, `yaw_lowpass_hz` 100, `acc_lpf_hz` 25, PG ids 544/554 unchanged). Its title still says 4.5; happy to reword if you would prefer.

The tunes cannot be merged the same way. `d_min_roll`/`d_min_pitch`, `dshot_idle_value` and `simplified_dmax_gain` were all renamed in 2025.12, so a single tune file would have to set both spellings and would throw errors on both versions. The tune presets therefore stay split, with the 2025.12 files covering 2025.12 and 2026.6.

### Include fixes in the tunes

- RC_LINK options now include `presets/2025.12/rc_link/*` instead of `presets/4.3/rc_link/*`. The 4.3 versions set `rc_smoothing_feedforward_cutoff`, which does not exist in 2025.12 or 2026.6 and errors when those options are ticked. The 4.5 tunes keep the 4.3 includes, where that parameter is still valid.
- `aos_ul5_tune` was including the 4.4 copy of its filters, which lacks the RPM filter weight options and pulls in the 4.3 filter defaults. It now uses the same file as its siblings.
- `aos_5.5_tune` now includes `aos_5.5_filters` rather than `aos_5_filters`, in the 4.4 and 4.5 trees as well as 2025.12. The two have identical `set` commands, so this is a naming fix only.
- `aos_cine80_tune` included `aos_7_filters` and `aos_ul8_tune` included `aos_ul7_filters`, copy-paste from the tunes they were derived from â€” the dedicated filter presets arrived alongside those tunes in #494 and #504 and were then never referenced. Both now use their own. This one does change the filtering applied to those two airframes, and is fixed in the 4.4 and 4.5 trees too, since the same mistake is in all of them.
- Net result: every AOS filter preset in the repo is now referenced by the tune for its own frame. The only tune using another frame's filters is `aos_hdzero_aio5_65mm`, which is a 65mm build with no separate frame filters.

I diffed the full CLI parameter list between the `4.5.5`, `2025.12.5` and `2026.6.1` tags (settings.c literals plus the `PARAM_NAME_*` macros in `fc/parameter_names.h`) to check all of the above. `node indexer/check.js` passes, every `INCLUDE` target resolves, and every AOS filter preset is now referenced by its own tune. `index.json` / `index_hash.txt` are untouched, since CI regenerates them.

### One thing for the maintainers

`presets/2025.12/tune/defaults.txt` is included by every AOS tune, and by other authors' 2025.12 tunes, and still sets five parameters that were removed in 2026.6:

- `abs_control_gain`, `abs_control_limit`, `abs_control_error_limit`, `abs_control_cutoff`
- `transient_throttle_limit`

Applying any of these tunes to 2026.6 will therefore report five CLI errors, even though the tune itself applies correctly. That is an `OFFICIAL` file so I have left it alone here, but it needs a 2026.6 pass before tune presets are error-free on the new release. Happy to raise a separate PR for it.


limonspb:
im a bit out of the loop. Tune/Filter default settings for 2026 need and update or no? @sugaarK

ChrisRosser:
@limonspb Yes, they need an update, and it is not an AOS-specific problem. I have raised #595 with the fix so there is something concrete to look at.

`presets/2025.12/tune/defaults.txt` is `STATUS: OFFICIAL`, declares `FIRMWARE_VERSION: 2025.12` only, and is included by **43 tunes across 8 authors** — aos_rc (21), uav_tech (12), supafly_fpv (4), karate (2), plus openracer, blackbird, basher and Funkolius.

Diffing `resetPidProfile()` between the `2025.12.5` and `2026.6.1` tags, there are seven differences, not the five I listed in the description above.

**Removed in 2026.6** — these report CLI errors on apply:

```
abs_control_gain, abs_control_limit, abs_control_error_limit,
abs_control_cutoff, transient_throttle_limit
```

**Default value changed** — I had missed these when I wrote the description:

| parameter | 2025.12 | 2026.6 |
|---|---|---|
| `d_max_gain` | 37 | 0 |
| `d_max_advance` | 20 | 35 |

That second pair matters more than the errors, and it changes the fix. I had assumed this was "drop the five dead lines and add `FIRMWARE_VERSION: 2026.6`". It is not: doing that would leave a preset whose stated job is *"Resets all Basic Tune category parameters to defaults"* silently writing two values that are not defaults on 2026.6. Nothing reports that, so it is worse than the five visible errors.

Since `INCLUDE` carries no version conditional, no single file can be correct for both releases. So #595 adds `presets/2026.6/tune/defaults.txt` as a new file, following the same one-per-version-folder convention as 4.3, 4.4, 4.5 and 2025.12, and leaves the 2025.12 file alone — it is correct for 2025.12, and stripping the `abs_control_*` resets there would stop it resetting parameters that still exist on that release.

To be clear about what #595 does **not** do: it does not repoint any tune at the new file, so nothing changes for users yet. Whether tunes on 2026.6 should reset to 2026.6 defaults is your call, and it is not free — a tune covering both releases can only carry one `INCLUDE`, so adopting it means splitting those tunes per version, the same way they already split at 4.5 → 2025.12 for the renamed `d_min_*` / `dshot_idle_value` / `simplified_dmax_gain` parameters. I would rather flag that cost than assume it. Happy to do the follow-up for my own presets, and to help with the others if those authors are willing.

None of this blocks this PR. #592 inherits the same pre-existing errors as every other 2025.12 tune, and no more.

---

On the two CodeRabbit threads still showing "Major" above — both ask me to repoint the filter includes at `presets/2025.12/filters/aos_rc/*`. That directory does not exist after this PR: those 18 files were byte-identical to their `presets/4.5` counterparts apart from the version line and an `OPIONS` typo, so they are deleted here and the `presets/4.5` originals now declare `4.5`, `2025.12` and `2026.6`. Following the suggestion would point every tune at files that are not there. I have left the threads for a human to close rather than resolving my own review comments.

`node indexer/check.js` passes and every `INCLUDE` target resolves.

PR #600, opened 2026-08-22
Add UAV Tech presets for Betaflight 2026.6

SupaflyFPV:
## Summary

Migrates the UAV Tech preset family to stable Betaflight 2026.6:

- 12 tune presets covering whoops through 10-inch/cinelifter builds
- Analog, DJI and DJI GPS OSD presets
- updated 2026.6 defaults and include paths
- normalized four-space option indentation

The presets are grouped by author to keep their tune range and matching OSD layouts together for review.

## Dependencies

Depends on #596 and #597. Until those foundation PRs merge, this draft also displays their commits; the diff will collapse automatically after they land.

## Validation

- `node indexer/check.js` returns `OK`
- `git diff --check` is clean
- generated index files are excluded


limonspb (approved):

sugaarK (approved):

haslinghuis (approved):

saturn-fpv:
Some options set thrust linear, and this 2026.6 tune sets the same value as the 2025.12 tune did. But didn't thrust_linear change in a way, that the equivalent should now be a lower value for the same result?
(Referencing this PR: https://github.com/betaflight/betaflight/pull/15226)
The same goes for the AOS presets and the Supafly presets, but in case I'm wrong about this I didn't want to add comments to all three.

